PR #1, opened 2017-03-01
Yay ! introducing Circles.

daita:
Reshare is still buggy and the UI needs small improvment; but this is Circles.

Because the core of nextcloud needs some patch, I will put a test server somewhere so you can ... test


daita:
only works on: https://github.com/nextcloud/server/tree/shares-circles

PR #48, opened 2017-04-26
Properly generate URLs

LukasReschke:
The other stuff won't work when installed within a subfolder :-)

Signed-off-by: Lukas Reschke <lukas@statuscode.ch>

### DIFF
--- a/js/circles.js
+++ b/js/circles.js
@@ -172,7 +172,7 @@
 				var result = {status: -1};
 				$.ajax({
 					method: 'PUT',
-					url: OC.generateUrl('/apps/circles/circles/' + circleid + '/share'),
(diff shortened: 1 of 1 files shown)


js/circles.js
```diff
@@ -172,7 +172,7 @@
 				var result = {status: -1};
 				$.ajax({
 					method: 'PUT',
-					url: OC.generateUrl(OC.linkTo('circles', 'circles/' + circleId + '/share')),
+					url: OC.generateUrl('/apps/circles/circles/' + circleid + '/share'),
```
nickvergessen: nope

LukasReschke: 🙈 Good catch!

PR #49, opened 2017-04-26
Escape HTML elements

LukasReschke:
Otherwise this leads to some XSS vectors here.

Signed-off-by: Lukas Reschke <lukas@statuscode.ch>

### DIFF
--- a/js/circles.app.elements.js
+++ b/js/circles.app.elements.js
@@ -212,8 +212,8 @@ var elements = {
 		$.each(exact, function (index, value) {
 			elements.membersSearchResult.append(
 				'<div class="members_search exact" searchresult="' +
-				value.value.shareWith + '">' + value.label + '   (' +
-				value.value.shareWith + ')</div>');
+				escapeHTML(value.value.shareWith) + '">' + escapeHTML(value.label) + '   (' +
+				escapeHTML(value.value.shareWith) + ')</div>');
 		});
 
 	},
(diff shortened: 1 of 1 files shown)


js/circles.app.elements.js
```diff
@@ -257,13 +257,12 @@ var elements = {
 	generateTmplCircle: function (entry) {
 		var tmpl = $('#tmpl_circle').html();
 
-		tmpl = tmpl.replace(/%title%/g, entry.name);
-		tmpl = tmpl.replace(/%type%/g, t('circles', entry.type));
-		tmpl = tmpl.replace(/%owner%/g, entry.owner.user_id);
-		tmpl = tmpl.replace(/%status%/g, t('circles', entry.user.status));
-		tmpl = tmpl.replace(/%level_string%/g, t('circles', entry.user.level_string));
-		tmpl = tmpl.replace(/%count%/g, entry.count);
```
LukasReschke: This seems unused and was not an object so I removed it as it failed otherwise.

js/circles.app.elements.js
```diff
@@ -272,12 +271,11 @@ var elements = {
 	generateTmplMember: function (entry) {
 		var tmpl = $('#tmpl_member').html();
 
-		tmpl = tmpl.replace(/%username%/g, entry.user_id);
-		tmpl = tmpl.replace(/%level%/g, entry.level);
-		tmpl = tmpl.replace(/%levelstring%/g, entry.level_string);
-		tmpl = tmpl.replace(/%status%/g, entry.status);
-		tmpl = tmpl.replace(/%joined%/g, entry.joined);
-		tmpl = tmpl.replace(/%note%/g, entry.note);
```
LukasReschke: This seems unused and was not an object so I removed it as it failed otherwise.

LukasReschke:
To test create a circle named `"><h1>asd<h1>asldj` and see that the HTML is not escaped.

PR #64, opened 2017-05-12
Import oc custom groups when possible

nickvergessen:
- [x] Create circles from custom groups
- [x] Create memberships
  - [x] Check if multiple admins is a problem code-wise @daita 
- [x] Create shares: share type = 1, share_with = customgroup_*


daita:
is there a way to detect who create the customgroup so we can define an owner with level=9 and others are moderator (level=6)

nickvergessen:
Not really, could only guess by order in the database.

nickvergessen:
Added the update of shares and only has one admin now.

PR #97, opened 2017-07-09
Spelling: ownership

comradekingu:


comradekingu:
If speaking of a subject, like a file, or some files, it is better to go with this. If all your rights are transferred, then "owner rights" is better imo.

The initial error was owner in owner right was written "Owner" in the middle of the sentence.

daita:
Thanks !

PR #98, opened 2017-07-11
Spelling: The member, non plural, of circle.

comradekingu:
This was initially a heads-up from javier.llamas on the translation platform about line 64.

"Successfully" also removed. If there is a non-successful way of doing things, or if it failed beforehand, i can reinstate it.

Being expelled/excluded from a circle rather than removed makes more sense for me, but i opted out of it since it is just more advanced for not a whole lot of reward, and potential problems in translation.

### DIFF
--- a/js/circles.app.results.members.js
+++ b/js/circles.app.results.members.js
@@ -61,7 +61,7 @@ var resultMembers = {
 
 				OC.dialogs.confirm(
 					t('circles',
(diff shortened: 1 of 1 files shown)


daita:
I know that the 'successful' thing is kind of redundant; let it like this and if you want to remove the word in the other notifications i will merge your modifications.

Expelled is nice.

Thanks for the work

PR #101, opened 2017-07-17
Spelling: into circle

comradekingu:
Consistent with https://github.com/nextcloud/circles/blob/3723f62ef930d73d59a5a62d637f9c59ef166f76/lib/Activity/Provider.php#L222


comradekingu:
@daita Now with less accidental.

PR #126, opened 2017-09-06
add the collaborator search plugin

blizzz:
* [ ] depends on https://github.com/nextcloud/server/pull/6328

for NC 13, to have the sharee endpoint still provide circle results


daita:
i'll have a look this week. Can you provide some info on how to register the search plugin ?

daita:
we would also need to update https://apps.nextcloud.com/schema/apps/info.xsd

daita:
not an issue as the old Sharees is still up & running !

Issue #129, opened 2017-09-07
Add circle hooks


dtygel:
We believe that Circles App should offer hooks for key events, like:

*  circle creation
*  circle destruction
*  join circle
*  leave circle
*  invite to circle
*  change circle settings (like rename, etc)
*  change member permissions in circle

We at https://github.com/coletivoEITA/ would like to implement these hooks and create a Pull Request. Is this a feature of interest?

viniciuscb:
Please ignore PR #130 . I've made a proposal in PR #131

daita:
@viniciuscb : is it fine to close this request ?

viniciuscb:
@daita Yes, now we have hooks in circles! Nice!

daita:
Obrigado, fika drete :+1:

PR #131, opened 2017-09-07
Add Circle hooks

viniciuscb:
This is a proposal for solving #129, add hooks so external applications can listen to circle operations.

Tested in tags/v0.12.4

### DIFF
--- a/lib/Service/CirclesService.php
+++ b/lib/Service/CirclesService.php
@@ -270,6 +270,8 @@ class CirclesService {
 			}
 
 			$this->circlesRequest->updateCircle($circle);
+
+			$this->eventsService->onSettingsChange($circle);
 		} catch (\Exception $e) {
 			throw $e;
 		}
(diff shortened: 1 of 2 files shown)


daita:
Bom dia @viniciuscb,

Thanks for your contribution, however it seems that the emit/hook system is slightly old and ugly. Can you try to use the EventDispatcher class ?

viniciuscb:
Hi, I will see how we can use EventDispatcher instead of OC_Hook.

viniciuscb:
@daita , Bom dia!

I've just submitted the alteration you have suggested. 

Um abraço!

daita:
Abraco man,

Should created a new branch instead of a fork as I could edit and add the dispatcher in the automatic DI/__construct()

please check https://github.com/nextcloud/circles/commit/a0b31ba7bb3270c7782d501ff710167c481420b5

PR #148, opened 2017-11-13
Spelling: Marked nouns, in-valid/sufficient

comradekingu:
Is it possible to make "This member is the owner of the circle" be "Owner of circle" ?
While different from the regular structure, I don't see what the extra verbiage adds.


daita:
Let's keep 'This member is the owner of the circle'

PR #151, opened 2017-11-13
escaped ', lowercase, within

comradekingu:


templates/navigate.php
```diff
@@ -382,11 +382,11 @@
 						<td><input type="checkbox" value="1" id="settings-link-files"></td>
 					</tr>-->
 					<tr id="settings-entry-link-auto">
-						<td class="left"><?php p($l->t('Automatically Accept Link Request')); ?>
+						<td class="left"><?php p($l->t('Accept Link Request Automatically')); ?>
```
comradekingu: If this is not the intended meaning, it should be "requests"

Issue #163, opened 2017-12-18
Enable audit by configuration


fgsl:
This feature allows to enable audit of actions for app circles.

The motivation for this change is documented in [forum](https://help.nextcloud.com/t/registry-of-sharing-information/24038).

It's a feature configurable. It doesn't change current behaviour, only adds audit of actions. It can be desirable feature for customers that want to have an wide control of user actions.

The following actions are audited:

1. user X creates circle Z;
2. user X removes circle Z;
3. user X changes name of circle Z to circle W;
4. member X is added to circle Y by user Z;
5. member X is removed from circle Y by user Z;
6. user X shares fiel/folder with circle Y;
7a. user X, that created circle Y, unshares file/folder with circle Y;
7b. user Z,that is member of circle Y, unshares file/folder with circle Y.

fgsl:
I change name of signals of share and unshare to be compatible with admin_audit and activity apps.

It's related to https://github.com/nextcloud/circles/issues/202.

fgsl:
I updated this change and added sending of signals of sharing and unsharing to Activity.

PR #164, opened 2017-12-18
Enable audit by configuration #163

fgsl:


fgsl:
As change stayed for a while without review, I wide coverage of actions and will submit again.

Issue #188, opened 2018-04-07
Circles Unable to be shared to When not a member


Shinrai:
When a user is not part of a circle they are unable to share to a circle. If they can view the circle from the apps/circles page they should be able to share to that circle.

daita:
Unfortunately, sharing to a circle should only be possible to members

Shinrai:
I disagree. If you can visibly see the circle you should be able to share to it. 

Unlike groups you can see a circle and it can be visible to you without you being a member of that circle.
Per the module itself:

> Anyone can find the circle and request an invitation; but only members will see who's in it and get access to shared items.

Art4:
I would like to see this happen. My use case is the following: My organisation has self-organized teams. Now the management should be able to share a folder/calendar with `Team_A` without being member of `Team_A`.

ypatios:
I am also for this feature. Circles have several setting options; this could be one more for users to decide if they want to offer users to share resources with a circle if they themselves aren't a member of that circle. It is undoubtedly a very common use case at our institution.

ArtificialOwl:
This will not be implemented, because it is out of scope and will generate so much technical debt.
I tried to work on a solution but it generate too many exception case in the flow of the code.

However, you can still:

- Create a new Circle (Circle A),
- Invite in 'Circle A' the Circle you want to be able to share,
- share to Circle A.

Users from the sub-circle will have access to files shared to Circle A.

Shinrai:
Man this is old.. funny I gave them a PR that fixed the issue and allowed it to work. Ran it on my own install and had no problems. Not sure what technical debt there was with it. Was quite a simple fix that I implemented on my own way back when. Not even sure what issues artificialowl was talking about I had it running with zero issues back then. They wanted an admin option to enable/disable this feature but at the time there wasn’t even an admin option to enable/disable sharing within a circle.

enjeck:
> This will not be implemented, because it is out of scope and will generate so much technical debt. I tried to work on a solution but it generate too many exception case in the flow of the code.

Perhaps we can make this clearer from the UI or document somehow so the behavior is not a suprise to the user

PR #189, opened 2018-04-07
Update CollaboratorSearchPlugin.php

Shinrai:
Fixes Issue #188.

With this change the user is able to share to any circle they have access to view.


Shinrai:
See Pull request #190 for an updated pull.

Issue #223, opened 2018-08-09
Allow complete inheritance of class ProviderParser


fgsl:
Class ProviderParser has three private methods: setSubject, setRichSubject and setParsedSubject.
If they can be changed to protected, it is possible reuse them in new parser classes.

Motivation is issue #163, that has to be refactored. For refactoring, I will combine #163 with #202, so audit of action will include registering of activities. For that, I am writing a new parser class and I want to avoid repeating code.

joshtrichards:
Fixed in #224

PR #224, opened 2018-08-09
Allow complete inheritance of class ProviderParser #223

fgsl:


daita:
don't really see the point, but whatever, no problem for me

Issue #227, opened 2018-08-22
adding user to circle causes sql error


jmcclelland:
The error is:

`'SELECT "m"."user_id", "m"."user_type", "m"."circle_id", "m"."level", "m"."status", "m"."note", "m"."joined" FROM "oc_circles_members" "m" WHERE (("m"."status" = ?) OR ("m"."status" = ?) OR ("m"."status" = ?)) AND ("m"."level" >= ?) AND ("m"."circle_id" = ?) ORDER BY "m"."joined" ASC' with params ["Member", "Invited", "Requesting", "Member", "xxxxxxxxx"]`

As you can see, it is using the string "Member" instead of a integer level.

I fixed it with:

```
--- /tmp/circles-latest/circles/lib/Service/CirclesService.php  2018-08-22 17:18:30.585997366 -0400
+++ lib/Service/CirclesService.php      2018-08-22 17:56:01.229009852 -0400
@@ -557,18 +557,19 @@
        public function checkThatCircleIsNotFull(Circle $circle) {
 
                $members = $this->membersRequest->forceGetMembers(
-                       $circle->getUniqueId(), Member::STATUS_MEMBER, true
+                       $circle->getUniqueId(), Member::LEVEL_MEMBER, true
                );
```

pdschandler:
Same issue here.

PR #228, opened 2018-08-28
adding user to circle causes sql error #227

fgsl:


daita:
nice catch !

wiswedel:
resolves #227

Issue #238, opened 2018-10-02
Exclusion of circle does not remove file sharing


fgsl:
### Steps to reproduce
1. Create a circle of any type
2. Add members
3. Load a file
4. Share this file to circle
5. Destroy the circle

### Expected behaviour
File sharing deleted

### Actual behaviour
File sharing remains in table as a lost node.

### Solution
Remove file sharing from circle owner when circle is destroyed.

fgsl:
Well, I think that I misinterpreted the contents of the table share.

I verified that some messages in app activity changed after exclusion of circles with file sharing.
Unsharing parsed messages were replaced by subject. 

Then, I created a bug when I thought that I had solved another one...

Because of this, I close this issue.

PR #239, opened 2018-10-02
Exclusion of circle does not remove file sharing #238

fgsl:


daita:
what was wrong ? :]

fgsl:
Well, I think that I misinterpreted the contents of the table share.

I verified that some messages in app activity changed after exclusion of circles with file sharing.
Unsharing parsed messages were replaced by subject.

Then, I created a bug when I thought that I had solved another one...

Because of this, I close this issue.

PR #256, opened 2019-02-21
typo in de translation

Binford3000:
corrected a typo in de translation


daita:
please use https://www.transifex.com/nextcloud/ to add or modify localization !

PR #299, opened 2019-05-06
honor user enumeration prevention of the share api

### LINKED ISSUE #152: Add option to disable user autocomplementation
The Circles App does not honour the "Allow username autocomplementation in share dialog ..." setting! 

If Circles is enabled and you add a new circle and click on "add a member" it will propose usernames to you. That could leak user information of other users. 

It would be great to add an option to disable this autocomplementation feature, like the opt-out autocomplementation option for the sharing dialog

immerda:
If user enumeration is disabled in nextcloud core, then circle can
be used to circumvent this measure. For example the adduser button
in the user facing circle UI allows enumeration of all registered
users.

This patch honors the choice made in
 'shareapi_allow_share_dialog_user_enumeration'
for deciding if auto completion should present partial results.
This is in line with other apps, such as webdav, which reuse this
configuration choice to disable user enumeration.

In case this preference is set, all partial results are removed
from the results.

Fixes #152

### DIFF
--- a/lib/Controller/MembersController.php
+++ b/lib/Controller/MembersController.php
@@ -27,6 +27,7 @@
 namespace OCA\Circles\Controller;
 
 use OCA\Circles\Model\Member;
+use OCA\Circles\Model\SearchResult;
 use OCA\Circles\Service\MiscService;
 use OCP\AppFramework\Http\DataResponse;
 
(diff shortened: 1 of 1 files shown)


daita:
I like your approach, just fixed few thing

PR #300, opened 2019-05-10
l10n de_DE: fix typo

zerwes:
small typo fix for de_DE


daita:
please use https://www.transifex.com/nextcloud/ to add or modify localization !

Issue #242, opened 2018-10-08
sharing with circle requires https


schiessle:
I just tried circles 0.15.1 with Nextcloud 14.0.1, created a personal circle and then tried to share a folder with the circle. Then I get a 403 with this error message: "cURL error 7: Failed to connect to localhost port 443: Connection refused (see http://curl.haxx.se/libcurl/c/libcurl-errors.html)"

But if I reload the browser the share is there. So it seems like we create the share correctly but at some point in the process we try to make a https request with curl to the local server which is setup without ssl on my development system.

daita:
yes, this is the broadcast of the shares, this should be improved, but you can:

`./occ config:app:set circles --value 1 allow_non_ssl_links` 
`./occ config:app:set circles --value 1 local_is_non_ssl`

fgsl:
@schiessle and @daita, if I understand, the scenario is development, because if it is production, it is recommended that server has a SSL certificate. Then, I think that if is debug mode is on, settings allow_non_ssl_links and local_is_non_ssl could be assumed as on too.

Related methods of OCA\Circles\Service\ConfigService could be changed so:

```
	public function getLocalAddress() {
		return (($this->isLocalNonSSL() || \OC::$server->getConfig()->getSystemValue('debug',false) ? 'http://' : '')
			   . $this->request->getServerHost();
	}
```
```
	public function isNonSSLLinksAllowed() {
		if ($this->allowedNonSSLLinks === -1) {
			$this->allowedNonSSLLinks = 
				(int)$this->getAppValue(self::CIRCLES_ALLOW_NON_SSL_LINKS);
		}
                $this->allowedNonSSLLinks = (int)($this->allowedNonSSLLinks || \OC::$server->getConfig()->getSystemValue('debug',false));
		return ($this->allowedNonSSLLinks === 1);
	}
```
What do you think about it?

wiswedel:
resolved by #322

PR #322, opened 2019-07-28
Update README.md

wiswedel:
added documentation for mandatory config flag in non-ssl environments (see #242)

Signed-off-by: Sascha Wiswedel <sascha.wiswedel@nextcloud.com>


wiswedel:
@daita Are there more config flags of this kind? I'd appreciate if you could document things like that.
Thx.

wiswedel:
@daita Are there more config flags of this kind? I'd appreciate if you could document things like that.
Thx.

daita:
I could not find anything relevent in the config flags, thanks for your work

PR #323, opened 2019-08-02
Simplify icons and improve contrast

skjnldsv:
| Before | After |
|:---------:|:------:|
|closed|closed|
|personal|personal|
|secret|secret|
|public|black_circle|

I also centered the circle so it looks balanced when displayed.

| Before | After |
|:---------:|:------:|
|Capture d’écran_2019-08-02_11-14-49|Capture d’écran_2019-08-02_11-15-08|

This still causes issues because we return the direct img url in the share result, so we cannot benefit from the svg invert mechanism. But this is a first step towards proper dark theme support


jancborchardt:
The size of the "Lock" and "Eye" icon inside the circle could be a tiny bit bigger, but it’s fine. :)

PR #324, opened 2019-08-06
Improve stroke and circles width of icons

skjnldsv:
Design is better, but contrast and visibility was still not ok according to w3c.
Now it's better

| Before | After |
|:---------:|:------:|
|Capture d’écran_2019-08-06_11-02-47|Capture d’écran_2019-08-06_11-02-55|


daita:
@jancborchardt of course, but not today :) Also, we will still need a UI for users to manage their 'groups'

skjnldsv:
> @jancborchardt of course, but not today :) Also, we will still need a UI for users to manage their 'groups'

Ideally we should re-use the same as the users management.
Let's discuss this in september :)

PR #326, opened 2019-08-20
shared by mail using personal token

daita:


daita:
This PR:

- create a new table in the database.
- generate one entry per accounts with mail related types (mail address and contact) when a file/folder is shared to a Circle.
- send an URL routed by the Circle app itself, with a different token for each mail address of the Circle.
- When a recipient clicl on the link+token, we check that the mail address related to the token is still in the Circle.
- If it is fine, the viewer will be redirected to the normal URL using the token from SharesByMail.

If a mail address or a contact is removed from the Circle, the token from SharesByMail is refreshed.

PR #338, opened 2019-10-05
Fix grammar, unnecessary escape in closed circle description

hoxia:
Change incorrect use of "it's", and remove unnecessary escaped apostrophe by changing "who\\'s in it" to "its membership".

Note that this change likely needs bulk edits (find-and-replace) across localization files. **Only en_GB was edited to demonstrate** the likely needed change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ This is the right option if you want to do recurrent sharing with the same list 
 Anyone can see the circle, can join the circle and access the items shared to the circle.
  
 - Joining a **Closed Circle** requires an invitation or a confirmation by a moderator.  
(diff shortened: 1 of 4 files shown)


hoxia:
Looking back at this, the purpose of this probably isn't clear. The escaped "who\'s in it" appears _with_ the escape in the UI. To fix this, either the escape should be corrected, or an apostrophe should be avoided. It's possible to word the string more clearly without the apostrophe anyway.

AndyScherzinger:
Hi  @hoxia 

Unfotunately I'll have to close the PR. The translations in the l10n folder are fully handled via transifex, so they woudl need to be fixed at https://app.transifex.com/nextcloud/nextcloud/translate/#en_GB/circles/485650950

PR #341, opened 2019-10-16
Return member list when fetching details with force parameter

juliushaertl:
Required for https://github.com/nextcloud/deck/pull/1293


daita:
want a backport to 15 ?

PR #361, opened 2019-12-05
Support 18

nickvergessen:


daita:
Thanks @nickvergessen I was about to do this migration later today :]
Let me finish en contact backend and i'll merge this one

Issue #367, opened 2020-01-22
NC18: Circles fails on db-schema-creation and -migration with PostgreSQL as backend.


linuxrrze:
Just updated to NC 18 (with some trouble).

When trying to activate circles 0,18.0 I only get:

An exception occurred while executing 'DROP INDEX "primary"': SQLSTATE[42704]: Undefined object: 7 ERROR: index "primary" does not exist

Maybe something like adding "IF EXISTS" could fix that?

EmJotGeh:
Same problem here.

drumdum:
I have the exact same error with postgre sql server. I am on a clean installed version of NC 18.

EmJotGeh:
Yes, i should say that i'm using Postgre too!

linuxrrze:
Me too

JuniorJPDJ:
The same problem.
I'm also using postgres.
Error in log is:

From docker log:
```
postgres_1   | 2020-01-28T16:45:23.898718928Z 2020-01-28 16:45:23.898 UTC [886] ERROR:  index "primary" does not exist
postgres_1   | 2020-01-28T16:45:23.898763755Z 2020-01-28 16:45:23.898 UTC [886] STATEMENT:  DROP INDEX "primary"
app_1        | 10.201.2.6 -  28/Jan/2020:16:45:19 +0000 "POST /index.php" 500
```

From nextcloud log:
```
Doctrine\DBAL\Exception\DriverException: An exception occurred while executing 'DROP INDEX "primary"': SQLSTATE[42704]: Undefined object: 7 ERROR: index "primary" does not exist

    /var/www/html/3rdparty/doctrine/dbal/lib/Doctrine/DBAL/DBALException.php - line 169:

    Doctrine\DBAL\Driver\AbstractPostgreSQLDriver->convertException("An exceptio ... t", Doctrine\DBA ... ]})

    /var/www/html/3rdparty/doctrine/dbal/lib/Doctrine/DBAL/DBALException.php - line 145:

    Doctrine\DBAL\DBALException::wrapException(Doctrine\DBA ... {}, Doctrine\DBA ... ]}, "An exceptio ... t")

    /var/www/html/3rdparty/doctrine/dbal/lib/Doctrine/DBAL/Connection.php - line 1012:

    Doctrine\DBAL\DBALException::driverExceptionDuringQuery(Doctrine\DBA ... {}, Doctrine\DBA ... ]}, "DROP INDEX \"primary\"")

[...]

PR #371, opened 2020-02-12
Fix PostgreSQL compatibility

tortuetorche:
Since Circles [v0.18.2](https://github.com/nextcloud/circles/tree/v0.18.2)

Reference: https://github.com/nextcloud/circles/issues/367#issuecomment-584068412

### DIFF
--- a/lib/Migration/Version0017Date20200202112903.php
+++ b/lib/Migration/Version0017Date20200202112903.php
@@ -32,6 +32,7 @@ use Closure;
 use Doctrine\DBAL\Schema\SchemaException;
 use Doctrine\DBAL\Types\Type;
 use OCP\DB\ISchemaWrapper;
+use OCP\IConfig;
 use OCP\IDBConnection;
 use OCP\Migration\IOutput;
 use OCP\Migration\SimpleMigrationStep;
@@ -45,12 +46,16 @@ class Version0017Date20200202112903 extends SimpleMigrationStep {
 	/** @var IDBConnection */
 	private $connection;
 
+	/** @var  IConfig */
+	protected $config;
 
 	/**
 	 * @param IDBConnection $connection
+	 * @param IConfig $config
 	 */
-	public function __construct(IDBConnection $connection) {
+	public function __construct(IDBConnection $connection, IConfig $config) {
 		$this->connection = $connection;
+		$this->config = $config;
 	}
 
 	/**
@@ -90,6 +95,7 @@ class Version0017Date20200202112903 extends SimpleMigrationStep {
 			'instance', 'string', [
 						  'notnull' => false,
 						  'length'  => 255,
+						  'default' => $this->config->getSystemValue('instanceid'),
 					  ]
 		);
 		$table->addColumn(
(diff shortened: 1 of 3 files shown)


lib/Migration/Version0017Date20200202112903.php
```diff
@@ -90,6 +95,7 @@ public function changeSchema(IOutput $output, Closure $schemaClosure, array $opt
 			'instance', 'string', [
 						  'notnull' => false,
 						  'length'  => 255,
+						  'default' => $this->config->getSystemValue('instanceid'),
```
nickvergessen:
```suggestion
						  'default' => '',
```

The instance id is definetly a wrong setting there. if something should be used, it's the url of the server which is used when doing federation. But I think it's not good to put that into the database structure

daita: Default must be an empty string. I'll check that later today

tortuetorche: I didn't found any references for the `instance` column in the code, only this https://github.com/nextcloud/circles/blob/v0.18.2/lib/Service/ConfigService.php#L430
I guess there are some missing parts to use this not null column...

tortuetorche: > 
> 
> Default must be an empty string. I'll check that later today

I see, but not null constraint with empty string isn't possible with PostgreSQL

nickvergessen: but not-null is false, so empty is allowed. But in that case it needs a default, which is missing here.

tortuetorche: Got it, I'll update this pull request

lib/Migration/Version0017Date20200202112903.php
```diff
@@ -90,6 +89,7 @@ public function changeSchema(IOutput $output, Closure $schemaClosure, array $opt
 			'instance', 'string', [
 						  'notnull' => false,
 						  'length'  => 255,
+						  'default' => '',
```
juliushaertl: Shouldn't this go into a new migration? Otherwise instances that have already created the table will stay not get a default set there.

tortuetorche: You're right, I'm not very familiar with Nextcloud migrations but I'll try to add a new migration instead of updating this one

tortuetorche: Done, but we can't use an empty string (`""`) as default value with the new migration, because it doesn't work with [`Doctrine\DBAL\Schema\Table::changeColumn()`](https://github.com/doctrine/dbal/blob/v2.10.1/lib/Doctrine/DBAL/Schema/Table.php#L320) method... You can check yourself if you want

nickvergessen: Wehn you put it into a new migration it would work, as it then drops and recreates it :-X

tortuetorche: @nickvergessen Sorry, I'm not sure to understand what you said...
But you can't drop the `instance` column, because it's part of the primary key.
I guess when the default value is `""`, `Doctrine\DBAL\Schema\Table::changeColumn()` think it's a null value, so the default value of the column is unchanged...
But with the `Doctrine\DBAL\Schema\Table::addColumn()` method it works, the value `""` isn't considered as a null value

tortuetorche: > But with the `Doctrine\DBAL\Schema\Table::addColumn()` method it works, the value `""` isn't considered as a null value

So I suppose, there is inconsistency between the `addColumn()` method and the `changeColumn()` method

Issue #363, opened 2019-12-16
Feature request: offer setting so that any circle creates a group


rasos:
Our use case is that we need to map circle to groups, which are being synced to LDAP. Writing of LDAP members and groups is now supported in NC 17. This would allow us to connect other services to LDAP or an IDM such as keycloak and have other services such as RocketChat use the same group structure like Circles with the same members. This way, Circle would become the first LDAP compliant self-management tool for liquid organisations.

See also #128 which goes in the other direction: offer an option to restrict those who may assign groups.

tortuetorche:
Hi @rasos,

I've working on this feature, see https://github.com/tortuetorche/circles/tree/patches
But I need to split my big commit, which also fix other issues, in multiple commits and send pull request for each of them.

Have a good day,
Tortue Torche

Issue #342, opened 2019-10-16
Add group doesn't work with member limit -1|0


genestouxguy:
Hi,

Apps version : 0.17.8
Nextcloud version : 17.0.0

I tried 3 ways :

1. param Maximum member to 0
2. param Maximum member to -1
3. param Maximum member to 100

For 1 & 2, I obtained the error :
`Group contains too many members`

For 3 the group added.

For information, the param by default (admin) for Max member is 300 and the groups contain 1 to 3 members for now.

Sorry for my english.

croessner:
I can confirm this.

NC 17.0.1 and NC 16.0.5 both with current latest Circle app installed

codejp3:
Same here. 
NC 17.01
Circles 0.17.10

s3r3ndipit:
Same here, even with 0 member inside the group trying to link with a circle.

NC 18.0.2
Circles 0.18.3

Issue #128, opened 2017-09-07
Non-admin users should not be able to add group(s) to circles


dtygel:
In Nextcloud, when an user belongs to a specific group, she cannot see other groups, even if she is admin of the group. If she is owner of the group, she can add new users, but these new users will forcibly belong to the group she is owner of. She cannot assign other groups to users she created.

With this behaviour in mind, users who are not global admins should not be able to add  groups to circles. Now this is possible in circles app. Isn't it a bug?

daita:
I will add:

- An option to limit the group linking to Global Admin only.
- A filter on the search of the groups to that it only display the group the Non-Global-Admin belongs to (if the option is checked)

Issue #360, opened 2019-12-03
add users without confirmation in a closed group


duub:
Why I can't add a user directly in a closed group? Is it possible to add a third option to the existing two?

1- invitation-acceptance
2- join-acceptance protocol 
3- an Admin adds a user without confirmation (same  functionality existing in secret groups)
I prefer to lose the first option, than the third... if simplicity is a preference...

duub:
I'm having a user case where i want to add users directly to a group and at the same time let them create shared links from the shared folder... So i would like to change the functionality here or #359

Art4:
In NC 17 you can go to Admin settings -> group ware to configure the circles app. There you can active the option "Skip invitation step to closed circles". I hope this will solve your issue.

joshtrichards:
Fixed in #416

PR #384, opened 2020-03-27
Multiple fixes and add Group Backend support

tortuetorche:
_NOTE: This big pull request will be eventually split in multiple pull requests..._

* Add `group_backend` option to mirroring the circle as a group (automatically suffixed by ` Circle`)
  You can customize prefix and suffix with the `group_backend_name_prefix` and `group_backend_name_suffix` options
  The main goal is to have better integrations with Nextcloud applications and clients. (fix https://github.com/nextcloud/circles/issues/363)
  You should also hide these mirrored circles in Nextcloud applications and clients, because their equivalent group will be used instead, via the `allow_listed_circles` option set to `0`
  * When the `group_backend` option is enabled, store the mirrored group id in the `group_id` column of the `circles_circles` table, useful when the circle's name is updated
  * Add option `allow_listed_circles` To show (`1`, default) or hide (`0`) circles in shared list, disabling this option is very useful when the `group_backend` option is enabled
  * Exclude `personal` and `secret` circles from Group backend, if the `group_backend` option is enabled
* Enhance `onSettingsChange` event to get old setting values (including `name` and `description` attributes).
* Add `allow_adding_any_group_members` option to allow (`1`, default) or not adding group members from any users. When value is set to `0` it's not possible anymore to add contacts as circle members for `private` and `public` circle types, and it's only possible for groups where the current user is a member or for global administrators. Because they can't be group members, useful when the `group_backend` option is enabled. There is already the `accounts_only` option but it applied for any types of circles... (fix https://github.com/nextcloud/circles/issues/128)
* You can now add members in `personal` circles when `skip_invitation_to_closed_circles` option is enabled (fix https://github.com/nextcloud/circles/issues/382 and https://github.com/nextcloud/circles/issues/360)
* Fix Parsing error in activities when user request to join a circle (see https://github.com/nextcloud/circles/issues/243)
* Fix CSS
* Fix translations
* Fix adding group with member limit `-1` or `0` (fix https://github.com/nextcloud/circles/issues/342)

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,26 @@ In non-SSL environments (like on development setups) it is necessary to set two 
 
 `./occ config:app:set circles --value 1 local_is_non_ssl`
 
+## Allow mirroring circles as groups
+
+```bash
+./occ maintenance:mode --on
+
+./occ config:app:set circles --value 1 group_backend # Mirroring circles as groups
+./occ config:app:set circles --value 0 allow_listed_circles # Hide circles in shared list, useful with the 'group_backend' option
+
+# ./occ config:app:set circles --value "🌀 " group_backend_name_prefix # You can customize group name prefix
+# ./occ config:app:set circles --value " " group_backend_name_suffix # Remove default group name suffix with a `space` character
+
+./occ config:app:set circles --value 12 allow_circles # Only show 'public' and 'closed' circles
+./occ config:app:set circles --value 1 skip_invitation_to_closed_circles
+
+./occ config:app:set circles --value 0 allow_files_filtered_by_circles # Disable files list filtering by circles in the 'files' application
+./occ config:app:set circles --value 0 allow_adding_any_group_members # Adding group members only for groups where the current user is a member or global administrators
+
+./occ maintenance:mode --off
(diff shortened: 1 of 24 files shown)


tortuetorche:
Hi @daita,

Can you tell me if I need to split this big pull request in one pull request per commit, please? 

Have a good day,
Tortue Torche

daita:
Hello,

Thanks for this huge commit, but I would prefer a list of small PR instead ! (also, some of those are already fixed ;p)

Regards,

tortuetorche:
Hi @daita,

> Thanks for this huge commit, but I would prefer a list of small PR instead ! (also, some of those are already fixed ;p)

It's done, here my pull requests: #390, #391, #392, #393, #394, #395, #396, #397, #398, #399, #400 and #401!

NB: You should merge them in the ascendant order, otherwise some issues might appears...

Have a good day,
Tortue Torche

joshtrichards:
Closing since this PR was subsequently handled through split up PRs: https://github.com/nextcloud/circles/pull/384#issuecomment-611456041

PR #392, opened 2020-04-09
Better application registering and events dispatching

tortuetorche:


appinfo/app.php
```diff
@@ -34,7 +34,4 @@
 
 
 $app = \OC::$server->query(Application::class);
-$app->registerNavigation();
-$app->registerFilesNavigation();
-$app->registerFilesPlugin();
-
```
daita, line 40: I would keep this in here, as it seems that the app.php is loaded only when the app itself is needed.

tortuetorche: During my tests, I saw the listeners (registered in this file) trigger twice each time, so this pull-request fix this issue.
And for the render method who merge all render methods in one, I borrowed the [Talk](https://github.com/nextcloud/spreed/blob/v8.0.7/appinfo/app.php#L24) application

nickvergessen: > as it seems that the app.php is loaded only when the app itself is needed.

This should not be the case and therefor it's fine when everything is done in a register method.
But it can be replaced soon for Nextcloud 20:
https://github.com/nextcloud/server/pull/20865

PR #396, opened 2020-04-09
Better 'creation_activity' option handling

tortuetorche:


daita:
As you saw on #387 the Activity Settings Page now implement a new 'circles_as_non_member' activity feed. 

Now, do we really need to keep the creation_activity option ?

daita:
As you saw on #387 the Activity Settings Page now implement a new 'circles_as_non_member' activity feed. 

Now, do we really need to keep the creation_activity option ?

tortuetorche:
> Now, do we really need to keep the creation_activity option ?

I don't know, I don't use this option, this pull request is just a refactoring of the existing code, to be more readable.

PR #398, opened 2020-04-09
Add `allow_adding_any_group_members` option

### LINKED ISSUE #128: Non-admin users should not be able to add group(s) to circles

tortuetorche:
To allow or not adding group members from any users, default to `1`.

When value is set to `0`, it's only possible to add groups where the current user is a member or for global administrators.

Fix https://github.com/nextcloud/circles/issues/128

### DIFF
--- a/lib/Search/LocalGroups.php
+++ b/lib/Search/LocalGroups.php
@@ -26,11 +26,25 @@
 
 namespace OCA\Circles\Search;
 
+use OC;
 use OCA\Circles\ISearch;
 use OCA\Circles\Model\Member;
 use OCA\Circles\Model\SearchResult;
+use OCP\IUser;
+use OCA\Circles\Service\ConfigService;
 
 class LocalGroups implements ISearch {
+
+	/** @var ConfigService */
+	private $configService;
+
+	/**
+	 * @param ConfigService $configService
+	 */
+	public function __construct(ConfigService $configService)
+	{
+		$this->configService = $configService;
+	}
 
 	/**
 	 * {@inheritdoc}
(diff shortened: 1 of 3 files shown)


lib/Search/LocalGroups.php
```diff
@@ -26,26 +26,48 @@
...
 	 * {@inheritdoc}
 	 */
 	public function search($search) {
 
 		$result = [];
-		$groupManager = \OC::$server->getGroupManager();
+		$groupManager = OC::$server->getGroupManager();
 
 		$groups = $groupManager->search($search);
+		$user = OC::$server->getUserSession()->getUser();
```
daita, line 58: should be added as a parameter of the `__construct()`. can you do the same with IGroupManager ?

lib/Service/MembersService.php
```diff
@@ -387,8 +389,18 @@ private function verifyIdentContact(&$ident, $type) {
 	 */
 	private function addGroupMembers(Circle $circle, $groupId) {
 
-		$group = OC::$server->getGroupManager()
-							->get($groupId);
+		$groupManager = OC::$server->getGroupManager();
+		$group = $groupManager->get($groupId);
+
+		$user = OC::$server->getUserSession()->getUser();
```
daita, line 395: should be added as a parameter of the `__construct()`. can you do the same with IGroupManager ?

lib/Service/MembersService.php
```diff
@@ -387,8 +389,18 @@ private function verifyIdentContact(&$ident, $type) {
...
+
+		if (!$this->configService->isAddingAnyGroupMembersAllowed() &&
+			$group instanceof IGroup && $user instanceof IUser &&
+			!$group->inGroup($user) && !$groupManager->isAdmin($user->getUID())
+		) {
```
daita, line 400: I am not a fan of this condition; let's assume that user is not null, I prefer having an unhandled exception than a failed condition. something like if _($group !== null && [...]->isAddingAnyGroupMembersAllowed() && not in group && not admin)_

PR #399, opened 2020-04-09
Add `allow_listed_circles` option

tortuetorche:
To list or not circles outside the Circles app.

Default value is `1`, to show circles in shared list.
Set it to `0` to hide them.


daita:
I do not understand the purpose of this PR. Please add some details !

Issue #421, opened 2020-04-14
No translation


Valdnet:
No translated descriptions:

- Member
- Moderator
- Admin
- Owner

![obraz](https://user-images.githubusercontent.com/47037905/79267188-078d0580-7e99-11ea-812d-7e5432155a42.png)

Translation of these descriptions in Transifex has been around for 3 years.

App:
Nextcloud 19.0.0 beta 2
Circles 0.18.5

Valdnet:
After checking the code, the non-translated descriptions are taken from this file:

https://github.com/nextcloud/circles/blob/afd92c4edf5e6d0fa85bb117cd87f80109eb194c/lib/Model/BaseMember.php#L410-L421

Valdnet:
I would also consider these descriptions.
Words:

- Local Member
- Group

are also missing for translation in Transifex. The other two words come from other functions.

https://github.com/nextcloud/circles/blob/afd92c4edf5e6d0fa85bb117cd87f80109eb194c/lib/Model/BaseMember.php#L430-L437

Valdnet:
Change
`$(this).find('.level').text(levelString);`
on
`$(this).find('.level').text(t('circles', levelString));`
in the file https://github.com/nextcloud/circles/blob/2d1314835e4e145a23f5f284705db1722ea6b2b9/js/circles.app.navigation.js#L409
The correct level names began to appear in your personal circles, but you still can't select the level to set for the user.

PR #422, opened 2020-04-16
Get translations for the user level

Valdnet:
Get translations for the user level in the personal circle.


Valdnet:
@daita I think it's worth taking a look at these PR before releasing the new version of Circle.
The problem described in the issue: #421.

PR #429, opened 2020-06-05
Fixing missing FR translation

AisukoWasTaken:
An entry was missing in the french translation : "Name of the Circle" which translates to "Nom du Cercle".
I added it in the french locale file.


nickvergessen:
Translations are handled via transifex, so this needs to be done via:
https://www.transifex.com/nextcloud/nextcloud/translate/#fr/circles

PR #447, opened 2020-07-06
fix: add fail function to SharesController

azul:
`SharesController` extends `Controller` - not `BaseController`
and thus does not inherit `this->fail`.

Not sure if it should extend `BaseController` instead.

This is the fix with the least changes
that make error reporting in the logs work for me again.


daita:
can you also duplicate `success()` ?

azul:
> can you also duplicate `success()` ?

done :white_check_mark:

PR #452, opened 2020-07-07
fix: do not rely on `api` global

azul:
The api files relied on an api global to handle responses:

```javascript
$.ajax({ ...
}).done(function (res) {
	api.onCallback(callback, res);
}).fail(function () {
	api.onCallback(callback, result);
});
```

In the files app this only worked
because circles.files.app.js added api as a global.

Remove that global and replace it with
```javascript
OCA.Circles.api
```
wherever it was used.

Signed-off-by: Azul <azul@riseup.net>

### DIFF
--- a/js/circles.v1.circles.js
+++ b/js/circles.v1.circles.js
@@ -29,26 +29,19 @@
 var circles = {
 
 	createCircle: function (type, name, callback) {
-
-		var result = {status: -1};
-		$.ajax({
+		OCA.Circles.api.request({
 			method: 'PUT',
 			url: OC.generateUrl('/apps/circles/v1/circles'),
 			data: {
 				type: type,
 				name: name
 			}
-		}).done(function (res) {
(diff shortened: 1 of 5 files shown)


azul:
Added another commit. While the first fixes the actual problem the second is a proposed refactoring to reduce duplication and handle all the jquery specifics in a single place.

azul:
@daita, do you have some cycles to review? Or who else could is ask?

azul:
@juliushaertl looks like i need at least two reviews. So adding you as well as you worked on circles lately.

daita:
Hello,

Code looks good, haven't time yet to test it

mejo-:
I can confirm that the fix works as intended, if that's of help :blush:

daita:
would it be possible to rebase your PR on master ?

azul:
@daita, sure... gimme a second.

PR #462, opened 2020-07-14
Include mail matches in the collaborator search

juliushaertl:
This will also include user type search results for matching email addresses.


daita:
any backports ?

PR #463, opened 2020-07-14
[stable18] Fix collaborator search and include email matches

juliushaertl:
Backport of https://github.com/nextcloud/circles/pull/462 and fix an issue where the displayname of search results doesn't get passed properly to the SearchResult instance


daita:
nice catch on the argument!

Issue #465, opened 2020-07-15
Add a Member function is not working


SLYtiger16:
Nextcloud version: 19.0.0
Circles version: 0.19.3

### Steps to Reproduce 
1.  Create a circle of any kind
2.  click a user from the add user drop down

### Expected Behavior
User is added to the circle

### Actual Behavior
After about 5-10 seconds the user is listed and a Green notification pops up saying the user has been added. Then hit refresh or change to another circle and come back and the user is not there anymore. Also if I try to adjust their role I get the error that the user does not exist.
**Adding user**
![image](https://user-images.githubusercontent.com/36513817/87606560-e79bf680-c6c0-11ea-9122-1239c35022ce.png)
**Changing Role**
![image](https://user-images.githubusercontent.com/36513817/87606619-0bf7d300-c6c1-11ea-9b43-ef6ff110dad7.png)
**After Page Refresh or navigation away and back**
![image](https://user-images.githubusercontent.com/36513817/87606658-2467ed80-c6c1-11ea-8df7-2fd2a6b7f556.png)

SLYtiger16:
Also - There are no logs and server restart has no effect.

daita:
In your `config/config.php` there should be a list of _trusted_domains_; is the first one from the list a valid domain (or IP) that is reachable from the https hosting your Nextcloud ?

gerryvaz:
@daita 
Changing the first entry in the config.php _trusted_domains_ block to the nextcloud domain resolved issue in my instance.

> 'trusted_domains' => 
> array (
> **_0 => 'nextcloud.example.com',_**
> ),


Thanks.

SLYtiger16:
@daita
Changing the first entry in the config.php trusted_domains block to the nextcloud domain Does Not resolve the issue in my instance.

'trusted_domains' =>
array (
0 => 'cloud.mysite.com',
1 => '192.168.xxx.xxx'
),

Thanks.

daita:
is https://cloud.mysite.com/ available locally from your httpd ?

Camui87:
I have the same problem, complete new instance of Nextcloud 19.0.1 with PHP7.4 and PostgreSQL12
I can create a circle and for a few seconds everything looks fine when members are added, but they weren't added at all. If a role is changed I also get the message that the user doenst exist.

But the trusted domain to the first entry in the array of trusted domains and the domain is also reachable from within dns of the server

ranti71:
[...]

PR #474, opened 2020-08-09
Make all instances of `Request()` aware of the `self_signed_cert` setting

mejo-:
Fixes: #465

### DIFF
--- a/lib/Command/MembersCreate.php
+++ b/lib/Command/MembersCreate.php
@@ -175,6 +175,9 @@ class MembersCreate extends Base {
 		}
 
 		$request = new Request('/users', Request::TYPE_GET);
+		if ($this->configService->getAppValue(ConfigService::CIRCLES_SELF_SIGNED) === '1') {
+			$request->setVerifyPeer(false);
+		}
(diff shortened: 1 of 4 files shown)


mejo-:
To be honest I wonder why the circles app backend uses `daita\MySmallPhpTools\Request` at all.

First, I'd expect the backend to do direct library calls instead of calling a HTTP API for actions on local NC instance, no?

And second, why not use the Request() class that NC brings, i.e. `OC\AppFramework\HTTP\Request`?

I'm pretty sure there's reasons why it's the way it is and I didn't dive too much into the app code yet, so maybe I'm just ignorant and didn't look enough :wink:

If you intend to keep using `daita\MySmallPhpTools\Model\Request`, maybe a cleaner solution would be to accept the `verifyPeer` bool as optional argument to its constructor, make `$this->configService->getAppValue(ConfigService::CIRCLES_SELF_SIGNED)` return a bool and pass it directly to the `Request()` constructor?

daita:
> First, I'd expect the backend to do direct library calls instead of calling a HTTP API for actions on local NC instance, no?

yes, I need to migrate this.

PR #519, opened 2020-11-12
new listeners

### LINKED ISSUE #515: Use typed event listener instead of old deprecated OC_Hooks
Used in https://github.com/nextcloud/circles/blob/master/lib/AppInfo/Application.php#L136

Please use:

* `OCP\Group\Events\GroupDeletedEvent` for `'OC_User', 'post_deleteGroup'`
* `OCP\User\Events\UserDeletedEvent` for `'OC_User', 'post_deleteUser'`

```php
use OCP\Group\Events\GroupDeletedEvent;
use OCA\Circles\Listener\ GroupDeletedListener;

...
$dispatcher = $this->get(IEventDispatcher::class);
$dispatcher->addServiceListener(GroupDeletedEvent::class, GroupDeletedListener::class);
```

Ideally this happens in the `register()` method of the application. (see https://github.com/nextcloud/server/blob/bd6a6cf3bff74dc5690fe032a4203165227f01d2/apps/user_status/lib/AppInfo/Application.php#L70 for an example)

See https://github.com/nextcloud/server/blob/caff1023ea72bb2ea94130e18a2a6e2ccf819e5f/apps/files_sharing/lib/Listener/UserAddedToGroupListener.php for example how the listener can look like.

Those hooks will be dropped in Nextcloud 21 and you can use the new system in all versions starting from Nextcloud 17.

daita:
should fix #515


lib/AppInfo/Application.php
```diff
@@ -87,6 +92,11 @@ public function __construct(array $params = array()) {
 	 * @param IRegistrationContext $context
 	 */
 	public function register(IRegistrationContext $context): void {
+		$dispatcher = $this->getContainer()
+						   ->get(IEventDispatcher::class);
+
+		$dispatcher->addServiceListener(UserRemovedEvent::class, UserDeleted::class);
+		$dispatcher->addServiceListener(GroupDeletedEvent::class, GroupDeleted::class);
```
MorrisJobke, line 99:
```suggestion
		$context->registerEventListener(UserRemovedEvent::class, UserDeleted::class);
		$context->registerEventListener(GroupDeletedEvent::class, GroupDeleted::class);
```

MorrisJobke (changes requested):
See inline comment

lib/AppInfo/Application.php
daita: fixed, and using UserDeletedEvent :]

daita:
Should be okay, and it's working

MorrisJobke (approved):

PR #546, opened 2021-01-18
new circle model

daita:


daita:
TODO: password on share might no be compatible with settings shared between instances

PR #551, opened 2021-01-25
Circles 22

daita:
### rewrite for 22

- stop providing 'types' of Circles, but instead you can flag a Circle to fit your needs, some flag are only available from 'system side', in case an app wants to generate some groups (cf. documentation) 
 - Nextcloud user got their own Id, so a Circle or a User can be considered the same (named FederatedUser) from the PoV of an app,
 - Federated Circles with different level of trust (cf. documentation),
 - Caching of memberships, as some Circles can be members of other Circles. 

### documentation

- PR: https://github.com/nextcloud/documentation/pull/6220
- built and available online at https://valentine.pr0npaganda.net/documentation/groupware/circles/index.html


daita:
rebased!

PR #552, opened 2021-01-25
ocs

daita:
some ocs sample to retrieve circles and members

- get all circles that user can access (can be a non-member if circle is visible):
```
curl -u test:test -X GET 'https://cloud.example.net/ocs/v1.php/apps/circles/circles/' -H "OCS-APIRequest: true"
```

- get members from a circle, using uniqueId:
```
curl -u test:test -X GET 'https://cloud.example.net/ocs/v1.php/apps/circles/circles/f40e34f63dc13b/members' -H "OCS-APIRequest: true"
```

### DIFF
--- a/lib/Controller/OcsController.php
+++ b/lib/Controller/OcsController.php
@@ -1,0 +1,124 @@
+<?php declare(strict_types=1);
+
+
+/**
+ * Circles - Bring cloud-users closer together.
+ *
+ * This file is licensed under the Affero General Public License version 3 or
+ * later. See the COPYING file.
+ *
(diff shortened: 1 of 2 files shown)


lib/Controller/OcsController.php
```diff
@@ -0,0 +1,132 @@
...
+use OCP\IRequest;
+use OCP\IUserSession;
+
+
+/**
+ * Class OcsController
+ *
+ * @package OCA\Circles\Controller
+ */
+class OcsController extends Controller {
```
rullzer, line 49: It should extend the OCS controller

daita:
locked on a bad branch, please use #555

PR #555, opened 2021-01-30
ocs (2)

daita:
some ocs sample to retrieve circles and members

- get all circles that user can access (can be a non-member if circle is visible):
```
curl -u test:test -X GET 'https://cloud.example.net/ocs/v1.php/apps/circles/circles/' -H "OCS-APIRequest: true"
```

- get members from a circle, using uniqueId:
```
curl -u test:test -X GET 'https://cloud.example.net/ocs/v1.php/apps/circles/circles/f40e34f63dc13b/members' -H "OCS-APIRequest: true"
```

### DIFF
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -25,8 +25,6 @@
  */
 
 return [
-	'ocs'    => [
-	],
 	'routes' => [
 		['name' => 'Settings#getSettings', 'url' => '/admin/settings', 'verb' => 'GET'],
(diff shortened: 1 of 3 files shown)


daita:
Adding a new route to obtain memberships:

```
curl -u test:test -X GET 'https://cloud.example.net/ocs/v1.php/apps/circles/users/{userId}/circles' -H "OCS-APIRequest: true"
```

daita:
rebased on stable21.

A huge rewrite of the apps come for NC22, with some modification in the OCS.
This will not be merged.

PR #566, opened 2021-03-10
Allow to get circle details with forceAll

juliushaertl:
This allows to get a circle details with the member list even if no current user is available (in a background job) in case the API consumer needs a full list of members, e.g. to issue notifications to them.

Should fix #563

PR #577, opened 2021-04-29
enforce forceAll

### LINKED ISSUE #576: Circles 0.21.1 with NC 21.0.1 and Calendar 2.2.0 Results in Calendar App Broken
Like the title says. I upgraded to version 0.21.1 today for the Circles app. I have NC 21.0.1 and Calendar 2.2.0. This results in the calendar being inaccessible from the browser. I cannot see events in the dashboard, and clicking on the calendar results in spinning wheel that never resolves. NextCloud throws errors on every refresh attempt. I could see the calendar still from the Windows app (configured to view NC), but I'm unsure what caching (if any) the Windows app uses.

Downgrading the Circles app to version 0.21.0 corrects this problem.

This should be a cleaned error log from one of the refresh events:
[err.txt](https://github.com/nextcloud/circles/files/6381112/err.txt)

daita:
Should fix #576

### DIFF
--- a/lib/Service/CirclesService.php
+++ b/lib/Service/CirclesService.php
@@ -284,7 +284,7 @@ class CirclesService {
 				|| $circle->getHigherViewer()
 						  ->isLevel(Member::LEVEL_MEMBER)
 			) {
-				$this->detailsCircleMembers($circle);
+				$this->detailsCircleMembers($circle, $forceAll);
 				$this->detailsCircleLinkedGroups($circle);
 				$this->detailsCircleFederatedCircles($circle);
 			}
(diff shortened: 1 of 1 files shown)


luckyhandler (approved):

luckyhandler:
I only checked the code and didn't test it.

acsfer:
Not working here. Only reverting https://github.com/nextcloud/circles/pull/566 fixes those issues.

Art4:
Please excuse me if I'm wrong but this PR don't do anything. (I neglect a possible speed advantage)
It just moves the `$forceAll === true` as the first condition in the if-statement.

juliushaertl:
@daita I pushed another commit in order to also pass the forceAll parameter to the method that fetches the members list, which fixes the issue on my test setup.

> Please excuse me if I'm wrong but this PR doesn't do anything. (I neglect a possible speed advantage)
It just moves the $forceAll === true to the first position in the if-condition.

It is not only a speed advantage but due to the OR operation if the first one matches the others will never get evaluated, so the `$circle->getHigherViewer()->isLevel()` which attempted to call on null then would never be executed if forceAll is used. Unfortunately it was only the first half of the fix :wink: 

Anyways testing of the two patches together is highly appreciated.

Art4:
@juliushaertl Thank you for the explanation.

armouredking:
Modifications do not fix for me, see #577 (I can't reopen bug).

PR #617, opened 2021-06-14
update name instead of displayName

daita:


mejo- (approved):
Great, I just tested and it works like expected. Ready to merge from my side.

PR #618, opened 2021-06-14
Minimal invasive sanitization for `sanitizedName`

mejo-:
Instead of stripping all non-ascii characters (which doesn't work well
with non-latin scripts), do minimal sanitization for `sanitizedName`
that can be used as filename and in URLs:

* Replace `/` with `-` to prevent directory traversal
* Remove characters that create problems on certain filesystems
* Remove leading spaces or dots to prevent hidden files

### DIFF
--- a/lib/Service/CircleService.php
+++ b/lib/Service/CircleService.php
@@ -529,7 +529,7 @@ class CircleService {
 
 		$i = 1;
 		while (true) {
-			$testSanitizedName = $baseSanitizedName . (($i > 1) ? '-' . $i : '');
+			$testSanitizedName = $baseSanitizedName . (($i > 1) ? ' (' . $i . ')' : '');
 
 			$test = new Circle();
 			$test->setSanitizedName($testSanitizedName);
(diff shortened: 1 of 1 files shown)


mejo-:
I went with adding ` (n)` similar to the `displayName` as that's Nextclouds default when enumerating in case of namespace conflicts.

lib/Service/CircleService.php
```diff
@@ -555,15 +555,20 @@ private function confirmSanitizedName(Circle $circle) {
...
+		// replacing instead of stripping seems the better tradeoff here
+		$sanitized = str_replace('/', '-', $name);
+
+		// remove characters which are illegal on Windows (includes illegal characters on Unix/Linux)
+		// see also \OC\Files\Storage\Common::verifyPosixPath(...)
+		/** @noinspection CascadeStringReplacementInspection */
+		$sanitized = str_replace(['*', '|', '\\', ':', '"', '<', '>', '?'], '', $sanitized);
+
+		// remove leading spaces or dots to prevent hidden files
+		$sanitized = preg_replace('/^[\. ]+/mu', '', $sanitized);
```
daita, line 568: might be useless if you replace L571: `return trim($sanitized, '. ');`

mejo-: The diference would be that dots (`.`) would be removed from the end of the string as well. While I agree that they're not particularly useful, I still would stick to the minimal technically necessary changes.

I don't have hard feelings here though :wink:

mejo-: Changed to your suggested solution now :)

mejo-:
:purple_heart:

PR #637, opened 2021-06-23
Bypass initiator from 'membership' using direct 'member'

ArtificialOwl:
The idea is to keep the basic request (getCircles & getMembers) working even if the table 'membership' is emptied, using direct link between initiator and the table 'member', instead of only using the table 'membership' to manage complete inheritance


ArtificialOwl:
- add more code to maintain,
- make some request more complex, 
- not sure about the usefulness of the hack

Let's keep the code in a PR if needed one day.

PR #648, opened 2021-06-28
keeping an illusion of consistency

ArtificialOwl:
#643


dartcafe:
That works for me out of the box as well. 👍

PR #783, opened 2021-08-13
allow to visit Single Circle

ArtificialOwl:
Will be useful when searching for users as entity

### DIFF
--- a/lib/Model/Probes/CircleProbe.php
+++ b/lib/Model/Probes/CircleProbe.php
@@ -49,6 +49,8 @@ class CircleProbe extends MemberProbe {
 
 	/** @var bool */
 	private $includeNonVisible = false;
+	/** @var bool */
+	private $visitSingleCircles = false;
 
 
 	/**
@@ -85,6 +87,7 @@ class CircleProbe extends MemberProbe {
 
 		return $this;
 	}
+
 
 	/**
 	 * @param bool $include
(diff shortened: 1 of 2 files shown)


ArtificialOwl:
@nextcloud/server-backend please review !

lib/Model/Probes/CircleProbe.php
```diff
@@ -126,6 +130,25 @@ public function includeBackendCircles(bool $include = true): self {
 	}
 
 
+	/**
+	 * @param bool $visit
+	 *
+	 * @return $this
+	 */
+	public function visitSingleCircles(bool $visit = true): self {
```
skjnldsv, line 138: A bit of documentation on what it does? :)

ArtificialOwl: So, the provisioning API for Circles (for NC23+) should be able to returns all entities like Circles and Circles generated from Nextcloud Groups. However, with the current status of the code, the Circles App have no way to browse the list of singleIds for Nextcloud Users (as Entities).

This methods visitSingleCircles() allow to add a parameter to the Probe to that the request will also returns singleIds for Nextcloud Users (as a visitor as the current viewer/initiator does not belongs to the Single Circle)

skjnldsv: No, I meant code documentation :wink: 
Your functions on this are missing docs headers :see_no_evil:

ArtificialOwl: i will write this docs in the next big PR about public interface

skjnldsv: Well, all functions should have a documentation header. It's not about being public or not :wink:

mejo-: @skjnldsv I agree that a short descriptive header is helpful to understand the meaning functions. Though this class is probably a bad example to start. Still, I added documentation headers for the functions: https://github.com/nextcloud/circles/pull/783/commits/369cfe5ed757acf463357b6ed8ad704863eed84c

mejo- (approved):

ArtificialOwl:
no need to backport

PR #784, opened 2021-08-13
fix definition on single circles

ArtificialOwl:
The current definition returns 'Nextcloud User owned by userId'.
This will only returns 'Nextcloud User'


ArtificialOwl:
@nextcloud/server-backend please review !

skjnldsv (approved):

juliushaertl (approved):

PR #790, opened 2021-08-20
emulate initiator on CircleJoin

ArtificialOwl:
fix an issue when trying to join a Circle as initiator is empty at the time of the request


ArtificialOwl:
@nextcloud/server-backend  needs to be pushed for nc22.1.1

ArtificialOwl:
Added emulation also when CircleLeave, to allow cancellation of a JoinRequest (not-yet-members).

Adding a CircleProbe with includeNonVisibleCircles to allow this process on non-visible Circles

blizzz (approved):

PR #792, opened 2021-08-23
owner of NO_OWNER should not have memberships cached

ArtificialOwl:
If we keep caching memberships of owner of NO_OWNER Circles, the owner will be returned on `getInheritedMembers()`.

This PR will fix this while keeping the sql request from failing when searching for existing Circles based on a FederatedUser used as Owner of a NO_OWNER Circle

### DIFF
--- a/lib/Db/CoreQueryBuilder.php
+++ b/lib/Db/CoreQueryBuilder.php
@@ -126,6 +126,7 @@ class CoreQueryBuilder extends NC22ExtendedQueryBuilder {
 				self::MEMBERSHIPS => [
 					self::CONFIG
 				],
+				self::DIRECT_INITIATOR,
 				self::INITIATOR => [
 					self::OPTIONS => [
 						'minimumLevel' => Member::LEVEL_MEMBER
@@ -1156,13 +1157,36 @@ class CoreQueryBuilder extends NC22ExtendedQueryBuilder {
 			)
 		);
 
+
+		$listMembershipCircleAlias = [$aliasMembership];
+		if ($this->getBool('initiatorDirectMember', $options, false)) {
+			try {
+				$aliasDirectInitiator = $this->generateAlias($alias, self::DIRECT_INITIATOR, $options);
+				$listMembershipCircleAlias[] = $aliasDirectInitiator;
+			} catch (RequestBuilderException $e) {
+			}
+		}
+
 		try {
 			$aliasMembershipCircle = $this->generateAlias($aliasMembership, self::CONFIG, $options);
+			$orXMembershipCircle = $expr->orX();
+			array_map(
+				function (string $alias) use ($orXMembershipCircle, $aliasMembershipCircle) {
+					$orXMembershipCircle->add(
+						$this->expr()->eq(
+							$alias . '.circle_id',
+							$aliasMembershipCircle . '.unique_id'
+						)
+					);
+				},
+				$listMembershipCircleAlias
+			);
+
 			$this->leftJoin(
 				$aliasMembership,
 				CoreRequestBuilder::TABLE_CIRCLE,
 				$aliasMembershipCircle,
-				$expr->eq($aliasMembership . '.circle_id', $aliasMembershipCircle . '.unique_id')
+				$orXMembershipCircle
 			);
 		} catch (RequestBuilderException $e) {
 		}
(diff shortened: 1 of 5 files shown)


ArtificialOwl:
This issue is minor and apps like Deck or Talk already filters on getUserType(). getUserType() from Owner in a no_owner Circles is 10000

While this can be backported to stable22, we can wait for next RC (instead of rushing for 22.1.1) and check if apps are really affected.

ArtificialOwl:
because it is needed by #793 a backport would be appreciated

ArtificialOwl:
@nickvergessen @nextcloud/server-backend

juliushaertl:
@ArtificialOwl Ready to be reviewed? Might be good to request review then and set a label :wink:

lib/Db/CoreQueryBuilder.php
```diff
@@ -1156,13 +1157,36 @@ public function leftJoinInitiator(
 			)
 		);
 
+
+		$listMembershipCircleAlias = [$aliasMembership];
+		if ($this->getBool('initiatorDirectMember', $options, false)) {
+			try {
+				$aliasDirectInitiator = $this->generateAlias($alias, self::DIRECT_INITIATOR, $options);
+				$listMembershipCircleAlias[] = $aliasDirectInitiator;
+			} catch (RequestBuilderException $e) {
```
blizzz, line 1166: I would prefer at least a code comment why an Exception is ignored even if it is a valid use case.

lib/Db/CoreQueryBuilder.php
```diff
@@ -1156,13 +1157,36 @@ public function leftJoinInitiator(
...
+			}
+		}
+
 		try {
 			$aliasMembershipCircle = $this->generateAlias($aliasMembership, self::CONFIG, $options);
+			$orXMembershipCircle = $expr->orX();
+			array_map(
```
blizzz, line 1173: if `$listMembershipCircleAlias` can be big, use foreach for it has a much better memory footprint. Otherwise ignore :)

lib/Db/CoreQueryBuilder.php
```diff
@@ -1264,13 +1288,27 @@ protected function limitInitiatorVisibility(string $alias): ICompositeExpression
 
 		$minimumLevel = $this->getInt('minimumLevel', $options);
 		$andXMember = $expr->andX();
-		$andXMember->add(
-			$this->orXCheckLevel($levelCheck, $minimumLevel)
+		$orXLevelCheck = $expr->orX();
+
+		array_map(
```
blizzz, line 1293: as above

blizzz (commented):
minor remarks (superficial review)

lib/Db/CoreQueryBuilder.php
ArtificialOwl: it can only contains two fields, the `memberships` table, or the `members` table.

lib/Db/CoreQueryBuilder.php
ArtificialOwl: In fact this Exception have to be ignored.

There is some method in this class that are cross-called to generate the result.
As an example, a `member` belongs to a `circle` that have an `member` as owner that is based on a `circle`. You can see the redundancy here when retrieving a members list of a circle.
This is an example, but the idea is that `generateAlias()` will check [self::$SQL_PATH](https://github.com/nextcloud/circles/blob/0e1cab95a16b3abb9545265443724e82f9cbc65b/lib/Db/CoreQueryBuilder.php#L85) to build the request and avoid infinite loop.

Should I still comment in the code ?

ArtificialOwl:
> 
> 
> @ArtificialOwl Ready to be reviewed? Might be good to request review then and set a label 😉

it's ready and have to be reviewed :) sorry for the tags

lib/Db/CoreQueryBuilder.php
nickvergessen: You shall not comment in/out the code, but add a comment why exactly this exception is catched

blizzz: Yes, please mention it in the code.

skjnldsv (approved):

blizzz (approved):

PR #793, opened 2021-08-23
fix notification when invited to a circle

ArtificialOwl:
in certain condition `getMemberById()` will only works if we look for direct membership

will require #792


ArtificialOwl:
@nickvergessen @nextcloud/server-backend

blizzz (approved):

skjnldsv (approved):

PR #797, opened 2021-08-24
fixing inheritance

ArtificialOwl:
Fixing naming inheritance, using `inheritance` instead of `link`


ArtificialOwl:
backport is not urgent as no one is using this feature (yet)

The idea is that the FederatedUser returns as `inheritedBy` should contains the link between the user and the Circle itself and the path explaining how it gained membership.

PR #801, opened 2021-08-25
Make getLink work

nickvergessen:
getLink always through so far? Even if you where a member and have a link.
Also the getLink never returned a link?


ArtificialOwl (approved):

ArtificialOwl:
was not supposed to come before nc23, so was not in an hurry to fix this one :)

also #794

PR #803, opened 2021-08-25
exception on non visible circle

ArtificialOwl:


ArtificialOwl:
@nextcloud/server-backend  need a quick review

This would just avoid having an error on the front-end when refusing an invitation to a circle

blizzz (approved):

PR #808, opened 2021-09-01
fix an issue with displayName and force check on file ownership migration

ArtificialOwl:
- do not rewrite displayName (not useful on NC20)
- force details when ran from cli

### DIFF
--- a/lib/ShareByCircleProvider.php
+++ b/lib/ShareByCircleProvider.php
@@ -450,12 +450,12 @@ class ShareByCircleProvider extends CircleProviderRequest implements IShareProvi
 			$name = $data['circle_alt_name'];
 		}
 
-		$data['share_with'] =
-			sprintf(
-				'%s (%s, %s) [%s]', $name,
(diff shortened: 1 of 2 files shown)


lib/ShareByCircleProvider.php
```diff
@@ -450,12 +450,12 @@ private function editShareEntry($data) {
 			$name = $data['circle_alt_name'];
 		}
 
-		$data['share_with'] =
-			sprintf(
-				'%s (%s, %s) [%s]', $name,
-				$this->l10n->t(Circle::TypeLongString($data['circle_type'])),
-				$this->miscService->getDisplayName($data['circle_owner'], true), $data['share_with']
-			);
+//		$data['share_with'] =
```
PVince81, line 453: this isn't actually the display name, you'd need to put it into "share_with_displayname" instead, if useful

PVince81 (approved):
:+1: fixes transfer ownership, tested on stable21 also

PR #815, opened 2021-10-04
[stable21] Don't overwrite share_with on read

PVince81:
When reading with getSharesBy, the display name is extended to contain
the circle id.

It seems that this was written into the wrong property and would
overwrite the id that existed already in "share_with".

This fixes the discrepancy by storing the computed value in
"share_with_displayname".

Partial fix for https://github.com/nextcloud/server/issues/29059, please refer to that ticket for more context.

@ArtificialOwl was this a mistake or will this fix cause side effects ?

- [ ] forward port to stable22 and master's deprecated provider ?

### DIFF
--- a/lib/ShareByCircleProvider.php
+++ b/lib/ShareByCircleProvider.php
@@ -450,7 +450,7 @@ class ShareByCircleProvider extends CircleProviderRequest implements IShareProvi
 			$name = $data['circle_alt_name'];
 		}
 
-		$data['share_with'] =
+		$data['share_with_displayname'] =
 			sprintf(
 				'%s (%s, %s) [%s]', $name,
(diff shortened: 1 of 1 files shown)


ArtificialOwl:
If it's only for the ownership migration, there is already this: that https://github.com/nextcloud/circles/pull/808
It was a test on stable20, merged, not released yet and not backported to stable21

Can you test it with the same confition on your side ?

> [ ] forward port to stable22 and master's deprecated provider ?

Na, we need another way to implement this for 22, my current best bet is to dispatch an event when transferring ownership, so that the app manage the ownership on the share, instead of core

PVince81:
closing in favor of https://github.com/nextcloud/circles/pull/808

PR #857, opened 2021-11-29
Changelog: add NC22 migration notice, fixes #856

### LINKED ISSUE #856: NC 22: Circles migration too slow
We recently upgraded a Nextcloud with a 4 digit number of circles and 5 digit number of users to NC22. After the migration, as documentend in the known issues, the Circle memberships did not work. Just for clarity: that means that many shared files, Decks boards etc are not available for users.

First, we ran

`occ circles:maintenance --level 4`

which is recommended in the known issues section of the forum to facilitate a faster migration of Circles. In our case the command ran for about 3 days and in the end the circles were not migrated.

When looking into the source code, we found that the `circles:sync` command has some parameters that seemingly do what we want:
```
sudo -u www-data php -f /opt/nextcloud/occ circles:sync --migration [--force-run]
```
We're still waiting for the script to finish, but the migration seems to be making visible progress now.

However, the speed is again very limited. Currently, it migrates about 2 memberships per second. The total number of memberships is a 6 digit number.

To sum things up: The speed of the migration is very slow for large Nextcloud installations and the documentation for the problem is lacking. If Circles are widely used in an organization with a large NC installation, then this can be a blocking issue for a migration to NC22.

svenseeberg:
I think it would be prudent to warn admins about the issues with the Circles migration.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,9 +8,23 @@ All notable changes to this project will be documented in this file.
 - Full rewrite of the app for Nextcloud 22
 - Type of Circles are gone, replaced by config flags  
 - first implementation of federated circles (2.0)
-- first implementation of new ExtendedQueryBuilder
-  
-  (changelog in progress)
+- first implementation of new ExtendedQueryBuilder  
+- Important: Circles and Circle memberships will be migrated with cron jobs.
+  However, this can take a long time with large databases. The process can
+  be speed up by executing the migrations manually. This can still take a
+  couple of days. To migrate the Circles and memberships, use `occ`:
+  ```bash
+  occ circles:sync --migration [--force]
+  occ circles:memberships --all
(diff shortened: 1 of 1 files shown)


ArtificialOwl:
out of curiosity, how many users/circles do you have for the migration process to take days ?

ArtificialOwl:
also, please sign your commit to fix DCO !

svenseeberg:
We have about 30k users in the Nextcloud and nearly 2k circles.

ArtificialOwl (approved):

svenseeberg:
Currently, the count is at 194225.

svenseeberg:
I added even more info to the change log. With this info I would be satisfied enough to close the issues #857 & #856.

ArtificialOwl (approved):

PR #866, opened 2021-12-16
enforce password on new share

ArtificialOwl:
So, the hard part was to keep the clear password on the master instance while creating shares on slaves (in global scale setup)

- When a new member is added, or a new share is created, the main instance will generate a list of Clean Passwords and their Hash. Hash will be shared to slaves to be used when generating the share token,
- Each slave will use a different hashed password for each members and inherited members of the circle. And will returns the list of generate share,
- Main instance get informations about the generate shares, and generate a mail per members and inherited members with the right password from the list.

- [x] test in large global scale env,
- [x] test in single instance with mass member add,
- [x] create password the same way that shared by mail,
- [x] check local settings for enforcing password, add also an option in the apps itself

### DIFF
--- a/lib/Listeners/Files/PreparingMemberSendMail.php
+++ b/lib/Listeners/Files/PreparingMemberSendMail.php
@@ -33,10 +33,14 @@ namespace OCA\Circles\Listeners\Files;
 
 use ArtificialOwl\MySmallPhpTools\Traits\Nextcloud\nc22\TNC22Logger;
 use ArtificialOwl\MySmallPhpTools\Traits\TStringTools;
-use Exception;
 use OCA\Circles\AppInfo\Application;
-use OCA\Circles\Events\AddingCircleMemberEvent;
+use OCA\Circles\Events\PreparingCircleMemberEvent;
+use OCA\Circles\Exceptions\FederatedItemException;
+use OCA\Circles\Exceptions\RemoteInstanceException;
+use OCA\Circles\Exceptions\RemoteNotFoundException;
+use OCA\Circles\Exceptions\RemoteResourceNotFoundException;
 use OCA\Circles\Exceptions\RequestBuilderException;
+use OCA\Circles\Exceptions\UnknownRemoteException;
 use OCA\Circles\Model\Member;
 use OCA\Circles\Service\ConfigService;
 use OCA\Circles\Service\ContactService;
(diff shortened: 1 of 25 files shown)


lib/Listeners/Files/PreparingMemberSendMail.php
```diff
@@ -0,0 +1,148 @@
...
+		}
+
+		$federatedEvent = $event->getFederatedEvent();
+		$clearPasswords = $federatedEvent->getInternal()->gArray('clearPasswords');
+		$hashedPasswords = $federatedEvent->getParams()->gArray('hashedPasswords');
+
+		foreach ($members as $member) {
+			if (($member->getUserType() !== Member::TYPE_MAIL
+				 && $member->getUserType() !== Member::TYPE_CONTACT)
+				|| array_key_exists($member->getSingleId(), $clearPasswords)
```
mejo-, line 135: Maybe add a comment here on what the conditions are for? From my understanding, it's "Only process members of TYPE_MAIL or TYPE_CONTACT, but ignore those who already have a password set", right?

lib/Service/ConfigService.php
```diff
@@ -329,16 +331,43 @@ public function contactsBackendType(): int {
 	 * @return bool
 	 * @deprecated
 	 * should the password for a mail share be send to the recipient
-	 *
 	 */
-	public function sendPasswordByMail() {
-		if ($this->isContactsBackend()) {
-			return false;
+	public function sendPasswordByMail(): bool {
+		return false;
```
mejo-, line 336: A config function that *always* returns `false`? What's the purpose of this? If it's just a leftover from the past, why not remove it altogether? This looks suspicious :thinking:

lib/Service/ConfigService.php
```diff
@@ -348,15 +377,7 @@ public function sendPasswordByMail() {
...
-		if ($this->isContactsBackend()) {
-			return false;
-		}
-
-		if ($circle->getSetting('password_enforcement') === 'true') {
-			return true;
-		}
-
-		return ($this->config->getAppValue('sharebymail', 'enforcePasswordProtection', 'no') === 'yes');
+		return false;
```
mejo-, line 380: Same here.

mejo- (approved):
My comments are rather cosmetic - feel free to ignore them if you're in a rush.

I have to admit that I don't fully understand all aspects of these code changes yet, but as far as I got it (from your explanations and reading the code), this looks good to me :+1:

lib/Service/ConfigService.php
ArtificialOwl: well, it is marked as deprecated ;-)

lib/Service/ConfigService.php
ArtificialOwl: same, deprecated. Those methods are called by some old files that needs to be cleaned. Should be another PR

lib/Listeners/Files/PreparingMemberSendMail.php
ArtificialOwl: comments added

PR #891, opened 2022-01-05
share with group members only

ArtificialOwl:


lib/Service/SyncService.php
```diff
@@ -299,6 +311,7 @@ public function syncNextcloudGroup(string $groupId): Circle {
 			$event = new FederatedEvent(SingleMemberAdd::class);
 			$event->setCircle($circle);
 			$event->setMember($member);
+			$event->getParams()->sBool('force_group_member', true);
```
PVince81, line 314: is `sBool` the equivalent of `setBool` ?

PVince81 (commented):
code looks ok, not sure if this is still a draft?

PR #900, opened 2022-01-11
Add OCI ci and make tests not silently failing

CarlSchwan:


ArtificialOwl:
- [ ] verify that Oracle create/migrate database correctly

CarlSchwan:
@ArtificialOwl I ended up disabling the Api tests as they need way too much adaption. I guess they need to be ported away from the Circles v1 API to the `CirclesManager`

The Controller tests now are running again and are passing the ci and it's better than nothing :)

ArtificialOwl:
Issue on php7.40-oci seems to come from the fact some some SQL table field names/aliases are longer than 30 chars

ArtificialOwl:
rebased

ArtificialOwl:
I will work on this, but won't be backported for next minor release

lib/Migration/Version0024Date20220203123902.php
```diff
@@ -124,6 +124,11 @@ public function changeSchema(IOutput $output, Closure $schemaClosure, array $opt
 			$table->setPrimaryKey(['token', 'instance']);
 		}
 
+		if ($schema->hasTable('circles_membership')) {
```
CarlSchwan, line 127: this needs to be moved in another migration :)

ArtificialOwl: fixed :]

lib/Migration/Version0024Date20220317190331.php
```diff
@@ -0,0 +1,71 @@
...
+
+class Version0024Date20220317190331 extends SimpleMigrationStep {
+
+
+	/**
+	 * @param IDBConnection $connection
+	 */
+	public function __construct(IDBConnection $connection) {
+	}
+
```
CarlSchwan, line 49:
```suggestion
```

ArtificialOwl:
rebased, adding a condition that there is no primarykey before creation.

ArtificialOwl (approved):

PR #903, opened 2022-01-12
Paginate list of Circles through OCS query parameters

### LINKED ISSUE #902: Circles list pagination
The OCS route `/ocs/v2.php/apps/circles/circles` returns all circles on the instance.

I have a production instance with nearly a thousand circles, so it's quite an expensive query to perform. I'd like a way to paginate the circles list to avoid such a query and response payload.

tcitworld:
Closes #902


lib/Controller/LocalController.php
```diff
@@ -411,10 +411,12 @@ public function memberRemove(string $circleId, string $memberId): DataResponse {
 	/**
 	 * @NoAdminRequired
 	 *
+	 * @param int|null $limit
+	 * @param int|null $offset
 	 * @return DataResponse
 	 * @throws OCSException
 	 */
-	public function circles(): DataResponse {
+	public function circles(int $limit = null, int $offset = null): DataResponse {
```
ArtificialOwl, line 419: I would go with `int $limit = -1, int $offset = 0`

lib/Controller/LocalController.php
```diff
@@ -423,6 +425,13 @@ public function circles(): DataResponse {
 				  ->filterBackendCircles()
 				  ->addDetail(BasicProbe::DETAILS_POPULATION);
 
+			if ($limit !== null) {
```
ArtificialOwl, line 428: and remove conditions

lib/Controller/LocalController.php
```diff
@@ -423,6 +425,13 @@ public function circles(): DataResponse {
...
+				$probe->setItemsLimit($limit);
+			}
+			if ($offset !== null) {
```
ArtificialOwl, line 431: (both)

ArtificialOwl:
Can you also add this in the `AdminController.php` ?

ArtificialOwl (approved):

tcitworld:
I wanted to use this for https://github.com/nextcloud/contacts/issues/2605, which doesn't really do BC with NC major versions, so I guess so, yes.

#919 #920

PR #912, opened 2022-01-21
Fix loosing memberships in low depth

ArtificialOwl:
fix 2 issues:

- update old owner membership on ownership transfer,
- update membership if inherited level is kept but inheritance path changed


ArtificialOwl:
Fix #905, but the database needs a refresh. This will be run in a weekly background job, not during migration.

Also, won't be included in next weeks release

mejo- (approved):

PR #916, opened 2022-01-27
returns empty array on empty param

ArtificialOwl:
Quick fix to avoid an exception when using dav client:

https://github.com/nextcloud/server/blob/5cd5245ca8b7b37156476ad128131ac5abc3891c/apps/dav/lib/Connector/Sabre/FilesReportPlugin.php#L302

the feature is gone but let's not crash even when not used.
The feature will be back for nc24.


mejo- (approved):

ArtificialOwl:
not needed in the end.

PR #931, opened 2022-02-22
block/force circle types

ArtificialOwl:
block and enforce circle types:

Enforce 'VISIBLE' flag to every circles managed by your users:
```
./occ config:app:set circles circle_types_force 8
```

Block 'FRIEND' flag to every circles managed by your users:
```
./occ config:app:set circles circle_types_block 128
```


mejo- (approved):
Looks good to me. If `ConfigService->verify()` and `CircleService->create()` are the only two places where the circle type may be set or changed (or changes are verified), then it should be good to go :+1:

PR #936, opened 2022-02-23
set member as INVITED only if not external

ArtificialOwl:
There is an issue currently where if Circle is set as INVITE when adding an external member.

This fix will accept memberships for external member without any need of confirmation from the said external member.

INVITE means "new member needs to confirm their membership (invitation)

### DIFF
--- a/lib/FederatedItems/SingleMemberAdd.php
+++ b/lib/FederatedItems/SingleMemberAdd.php
@@ -411,7 +411,9 @@ class SingleMemberAdd implements
 		} catch (MemberNotFoundException $e) {
 			$member->setId($this->token(ManagedModel::ID_LENGTH));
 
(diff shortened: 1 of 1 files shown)


mejo-:
Mh, this may lead to confusion, no? If users set `CFG_INVITE` they probably don't expect external users to be added instantly when adding `TYPE_MAIL` or `TYPE_CONTACT`. Maybe it would be good to document this behaviour in the UI? What do you think?

ArtificialOwl:
This is how it was working before. Later, we can fix/add a feature that send a mail with an invitation instead but, at least, this patch fixes a blocking issue that makes external members not been added to a circle with `CFG_INVITE`

mejo-:
I see, makes sense. Good to do then.

mejo- (approved):

Issue #901, opened 2022-01-11
Oracle database compatibility


CarlSchwan:
## SQL in statements don't support more than 1000 entries

Issue found in:

- [ ] Db\CircleProviderRequestBuilder::limitToFiles
- [ ] Db\CircleProviderRequestBuilder::limitToCircles
- [ ] artificial-owl/my-small-php-tools/lib Db/Nextcloud/nc22/NC22ExtendedQueryBuilder::exprLimitInArray

## Comparaison with CLOB (==, group by)

- [ ]  https://github.com/nextcloud/circles/blob/master/lib/Db/CircleRequestBuilder.php#L80

ArtificialOwl:
all done

PR #950, opened 2022-03-09
Lighter SQL requests and compat with Oracle

ArtificialOwl:


ArtificialOwl:
fix #901
fix #789 

This is a huge PR, split in multiple commit:

1. **import of MSPT**: the external tools lib is now fully copied within the app, using `OCA\Circles\Tools` as namespace. 
2. **switch import to OCA\Circles\Tools**: calls to MSPT have been changed to reflect the import of the tools lib.
3. **population stored as setting**: the idea here is to remove the calculation of the population of Circles during the SELECT request. Instead, the population is stored as a Circle's Setting. The calculation will be added in another Pull Request, until then population will be displayed at 0.
4. **removing useless SELECT DISTINCT and GROUPBY**.
5. **instance can be empty string**: Oracle does not like `$expr->eq()` on empty string, switching to `$expr->emptyString()` when needed.

lib/Controller/RemoteController.php
```diff
@@ -534,4 +534,76 @@ public function exceptionResponse(
...
+	 *      return new DataResponse(['test' => 42]);
+	 * } catch (JsonNotRequestedException $e) {}
+	 *
+	 *
+	 * @throws NotLoggedInException
+	 * @throws JsonNotRequestedException
+	 */
+	private function publicPageJsonLimited(): void {
+		if (!$this->jsonRequested()) {
+			if (!OC::$server->get(IUserSession::class)
```
CarlSchwan, line 553: Dependency injection

lib/Controller/RemoteController.php
```diff
@@ -534,4 +534,76 @@ public function exceptionResponse(
...
+	}
+
+
+	/**
+	 * @param array $needles
+	 *
+	 * @return bool
+	 */
+	private function areWithinAcceptHeader(array $needles): bool {
+		$request = OC::$server->get(IRequest::class);
```
CarlSchwan, line 583: Dependency injection

lib/Controller/RemoteController.php
```diff
@@ -534,4 +534,76 @@ public function exceptionResponse(
...
+		foreach ($accepts as $accept) {
+			if (in_array($accept, $needles)) {
+				return true;
+			}
+		}
+
+		return false;
+	}
+
+	/**
```
CarlSchwan, line 595: php doc comments that just repeat the type information of the method are useless

CarlSchwan (approved):

PR #953, opened 2022-03-10
adding options 'is:?' to search

ArtificialOwl:
adding options is:member, is:moderator, is:owner, is:admin to the search bar to filter Circles

### DIFF
--- a/lib/Service/SearchService.php
+++ b/lib/Service/SearchService.php
@@ -31,27 +31,46 @@ declare(strict_types=1);
 
 namespace OCA\Circles\Service;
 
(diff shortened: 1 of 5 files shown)


lib/Service/SearchService.php
```diff
@@ -31,27 +31,46 @@
...
 		FederatedUsers::class
 	];
 
 
+	/** @var IURLGenerator */
+	private $urlGenerator;
+
+	/** @var CircleService */
+	private $circleService;
+
```
CarlSchwan, line 62:
```suggestion
	private IURLGenerator $urlGenerator;
	private CircleService $circleService;
```

CarlSchwan (approved):

PR #960, opened 2022-03-13
23.1.0

ArtificialOwl:


CarlSchwan:
composer is not happy :(

ArtificialOwl:
seems that `psr/container` needed a downgrade

PR #972, opened 2022-03-17
Cache getSharedWith() result

ArtificialOwl:
The idea is to cache the result from getSharedWith() which can be requested multiple times (at least 4) when loading a new page on Files App. 

Each cache is identified by a key based on:

- the singleId of the initiator, 
- the nodeId of the request, can be `0`
- the checksum of the `Probe` object.

The cache related to the singleId of the membership is reset when membership is updated.
All the cache is clear when creating/editing/deleting a share. _(It might be interesting to filter the memberships affected by the share)_

### DIFF
--- a/lib/Listeners/Files/MembershipsRemoved.php
+++ b/lib/Listeners/Files/MembershipsRemoved.php
@@ -31,7 +31,6 @@ declare(strict_types=1);
 
 namespace OCA\Circles\Listeners\Files;
 
-use OCA\Circles\Tools\Traits\TStringTools;
 use OCA\Circles\CirclesManager;
 use OCA\Circles\Db\ShareWrapperRequest;
 use OCA\Circles\Events\MembershipsRemovedEvent;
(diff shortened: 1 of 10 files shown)


lib/Listeners/Files/MembershipsRemoved.php
```diff
@@ -66,8 +67,8 @@ class MembershipsRemoved implements IEventListener {
 	/** @var CirclesManager */
 	private $circlesManager;
 
-	/** @var ShareWrapperRequest */
-	private $shareWrapperRequest;
+	/** @var ShareWrapperService */
+	private $shareWrapperService;
```
CarlSchwan, line 71:
```suggestion
	private ShareWrapperService $shareWrapperService;
```

CarlSchwan (approved):
Looks good

ArtificialOwl:
squashed

icewind1991:
Caching the shares also caches the file metadata of the root of the share. Breaking sync as the old etag will be reported after a change for the duration of the cache ttl.

PR #975, opened 2022-03-25
bypass circle types

ArtificialOwl:
This PR allow to assign an entity that can bypass force/block rules on Circles' types.

The entity is a singleId and can be a group, a circle or a single user

```
./occ config:app:set circles bypass_circle_types --value <singleId>
```


ArtificialOwl:
replaced by #976

PR #976, opened 2022-03-25
bypass/limit permissions

ArtificialOwl:
(replacement for #975)

This PR allow to assign different entities to limit/bypass some permissions. Each settings is defined by a singleId. Only members (and inherited) that belongs to the Entity defined by its singleId will be affected:

- Circle Creation via OCS API. This settings will limit the creation of a new circle only to members of the assigned Entity,
- Bypass Circles' types: when set, members of the assigned Entity will be able to bypass the limitation defined by `circle_types_block` and `circle_types_force`

```
./occ config:app:set circles limit_circle_creation --value <singleId>
./occ config:app:set circles bypass_circle_types --value <singleId>
```

### DIFF
--- a/lib/Service/CircleService.php
+++ b/lib/Service/CircleService.php
@@ -101,6 +101,9 @@ class CircleService {
 	/** @var MemberService */
 	private $memberService;
 
+	/** @var PermissionService */
+	private $permissionService;
+
 	/** @var ConfigService */
 	private $configService;
 
@@ -114,6 +117,7 @@ class CircleService {
 	 * @param FederatedUserService $federatedUserService
 	 * @param FederatedEventService $federatedEventService
 	 * @param MemberService $memberService
+	 * @param PermissionService $permissionService
 	 * @param ConfigService $configService
 	 */
 	public function __construct(
@@ -125,6 +129,7 @@ class CircleService {
 		FederatedUserService $federatedUserService,
 		FederatedEventService $federatedEventService,
 		MemberService $memberService,
+		PermissionService $permissionService,
 		ConfigService $configService
 	) {
 		$this->l10n = $l10n;
@@ -135,6 +140,7 @@ class CircleService {
 		$this->federatedUserService = $federatedUserService;
 		$this->federatedEventService = $federatedEventService;
 		$this->memberService = $memberService;
+		$this->permissionService = $permissionService;
 		$this->configService = $configService;
 
 		$this->setup('app', Application::APP_ID);
@@ -197,7 +203,7 @@ class CircleService {
 		}
 
 		$this->confirmName($circle);
-		$this->configService->confirmAllowedCircleTypes($circle);
+		$this->permissionService->confirmAllowedCircleTypes($circle);
 
 		$member = new Member();
 		$member->importFromIFederatedUser($owner);
--- a/lib/Service/PermissionService.php
+++ b/lib/Service/PermissionService.php
@@ -1,0 +1,184 @@
+<?php
+
+declare(strict_types=1);
+
+
+/**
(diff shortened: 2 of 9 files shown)


lib/Service/PermissionService.php
```diff
@@ -0,0 +1,186 @@
...
+		$circle->setConfig($config);
+	}
+
+
+	/**
+	 * @param int $bitwise
+	 *
+	 * @return array
+	 */
+	public function extractBitwise(int $bitwise): array {
```
CarlSchwan, line 173:
```suggestion
	/**
	 * @return int[]
	 */
	private function extractBitwise(int $bitwise): array {
```

lib/Service/PermissionService.php
```diff
@@ -0,0 +1,186 @@
...
+	}
+
+
+	/**
+	 * @param Circle $circle
+	 *
+	 * @return bool
+	 * @throws RequestBuilderException
+	 */
+	public function canBypassCircleTypes(Circle $circle): bool {
```
CarlSchwan, line 115:
```suggestion
	private function canBypassCircleTypes(Circle $circle): bool {
```

lib/Service/PermissionService.php
```diff
@@ -0,0 +1,186 @@
...
+
+
+	/**
+	 * @param string $config
+	 *
+	 * @throws InsufficientPermissionException
+	 * @throws RequestBuilderException
+	 * @throws InitiatorNotFoundException
+	 */
+	public function confirm(string $config): void {
```
CarlSchwan, line 93:
```suggestion
	private function confirm(string $config): void {
```

lib/Service/CircleService.php
```diff
@@ -101,6 +101,8 @@ class CircleService {
 	/** @var MemberService */
 	private $memberService;
 
+	private $permissionService;
```
CarlSchwan, line 104:
```suggestion

	/** @var PermissionService */
	private $permissionService;
```

lib/Service/PermissionService.php
```diff
@@ -0,0 +1,186 @@
...
+	/**
+	 * @param Circle $circle
+	 *
+	 * @return bool
+	 * @throws RequestBuilderException
+	 */
+	public function canBypassCircleTypes(Circle $circle): bool {
+		try {
+			if (!$circle->hasInitiator()) {
+				throw new MembershipNotFoundException();
```
CarlSchwan, line 118:
```suggestion
				return false;
```

Not sure if this is simpler...

lib/Service/PermissionService.php
```diff
@@ -0,0 +1,186 @@
...
+		if ($singleId === '') {
+			return;
+		}
+
+		$this->federatedUserService->mustHaveCurrentUser();
+		$federatedUser = $this->federatedUserService->getCurrentUser();
+		try {
+			$federatedUser->getLink($singleId);
+		} catch (MembershipNotFoundException $e) {
+			throw new InsufficientPermissionException();
```
CarlSchwan, line 104:
```suggestion
			throw new InsufficientPermissionException($this->l10n->t('You have no permission to create a new circle'));
```

And then you don't need the try catch in confirmCircleCreation. This also simplifies backtraces

And then

ArtificialOwl: the error message won't fit if i want to create a new value to be checked :(

lib/Service/PermissionService.php
ArtificialOwl: initiator must exist, i think it would be better to keep the exception

ArtificialOwl:
squashed, rebased

mejo- (approved):

PR #980, opened 2022-03-29
+runAsSuperSession

ArtificialOwl:
The idea is to run a single command from CirclesManager in a super session:

```
		/** @var CirclesManager $manager */
		$manager = \OC::$server->get(CirclesManager::class);
		$result = $manager->runAsSuperSession('getCircles', [new CircleProbe()]);
```

### DIFF
--- a/lib/CirclesManager.php
+++ b/lib/CirclesManager.php
@@ -63,6 +63,7 @@ use OCA\Circles\Service\FederatedUserService;
 use OCA\Circles\Service\MemberService;
 use OCA\Circles\Service\MembershipService;
 use OCA\Circles\Tools\Exceptions\InvalidItemException;
+use Throwable;
 
 /**
  * Class CirclesManager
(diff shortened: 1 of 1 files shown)


lib/CirclesManager.php
```diff
@@ -206,6 +207,40 @@ public function startAppSession(string $appId, int $appSerial = Member::APP_DEFA
 		$this->federatedUserService->setLocalCurrentApp($appId, $appSerial);
 	}
 
+	
+	/**
+	 * Run a single method as Super, then Super level will be removed
+	 *
+	 * @throws Throwable
+	 * @noinspection PhpUndefinedVariableInspection
```
CarlSchwan, line 215: This sounds scary, undefined variables will throw errors in PHP 8.2

ArtificialOwl: Returned is mixed
let's put this in standby, merge is not necessary.

PR #974, opened 2022-03-24
limit some feature when Circles is managed by an app

ArtificialOwl:
- Allow an App to gain control over a Circle:
```
		/** @var CirclesManager $circleManager */
		$circleManager = \OC::$server->get(CirclesManager::class);
		$circleManager->startSuperSession();
		$circleManager->flagAsAppManaged('tnxgB4xdBpgew6H4l9CJXEUauUNql71');
		$circleManager->stopSession();

```

- When marked as AppManaged, the Circle cannot be removed from the OCS API (ie. Contacts App)

- Allow an App to create its own Entity:
```
		/** @var CirclesManager $circleManager */
		$circleManager = \OC::$server->get(CirclesManager::class);
		$circleManager->startAppSession('my app', 11010);

		$circle = $circleManager->createCircle(
			'Test Circle',
			$circleManager->getLocalFederatedUser('test1')
		);
```
_identifying as an App does not give more rights. To have rights over a Circle, the Entity of the App needs to be member/moderator/admin of the targetted Circle._

- Also cleaning some bitwise operations.

PR #984, opened 2022-03-31
[stable22] limit some feature when Circles is managed by an app

ArtificialOwl:
manual backport of #974


mejo- (approved):
Not sure why the CI fails, but apart from that lgtm :)

PR #1000, opened 2022-04-06
remove child shares

ArtificialOwl:
1000!

Issue:

- userA shares a first file to a Circle (including userA and userB),
- userB leaves the share,
- Circle is destroyed, a new Circle is created (including userA and userB)
- Sharing the same file.

Fix:

- split getUserShares from a single user to a circle and getAllShares to a circle
- when a circle is deleted, or when a member is removed, get affected shares ids
- get orphans on a maintenance process
- delete all shares and child based on ids.
 

The request that get all orphan shares, based on `parent` `not null` and entry with `id`=`parent` exists:
```
SELECT `sh`.`id`, `sh`.`share_type`, `sh`.`share_with`, `sh`.`uid_owner`, `sh`.`uid_initiator`, `sh`.`parent`, `sh`.`item_type`, `sh`.`item_source`, `sh`.`item_target`, `sh`.`file_source`, `sh`.`file_target`, `sh`.`permissions`, `sh`.`stime`, `sh`.`accepted`, `sh`.`expiration`, `sh`.`token`, `sh`.`mail_send` FROM `oc_share` `sh` LEFT JOIN `oc_share` `p` ON `p`.`id` = `sh`.`parent` WHERE (`sh`.`share_type` = 7) AND (`sh`.`parent` IS NOT NULL) AND (`p`.`id` IS NULL)
```

The request that drop shares and child based on an array of ids:
```
DELETE FROM `oc_share` WHERE (`share_type` = 7) AND ((`id` IN ('6')) OR (`parent` IN ('6')))
```

### DIFF
--- a/lib/Db/ShareWrapperRequest.php
+++ b/lib/Db/ShareWrapperRequest.php
@@ -431,11 +431,56 @@ class ShareWrapperRequest extends ShareWrapperRequestBuilder {
 	 * @param string $initiator
 	 */
 	public function deleteSharesToCircle(string $circleId, string $initiator = ''): void {
-		$qb = $this->getShareDeleteSql();
-		$qb->andWhere($qb->exprLimit('share_with', $circleId));
+		$qb = $this->getShareSelectSql();
+		$qb->limit('share_with', $circleId);
 		if ($initiator !== '') {
 			$qb->limit('uid_initiator', $initiator);
 		}
+
+		$ids = array_map(
+			function (ShareWrapper $share): string {
+				return $share->getId();
+			},
+			$this->getItemsFromRequest($qb)
+		);
+
+		$this->deleteSharesAndChild($ids);
+	}
+
+
+	public function removeOrphanShares(): void {
+		$qb = $this->getShareSelectSql();
+		$expr = $qb->expr();
+		$qb->leftJoin(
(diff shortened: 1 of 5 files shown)


lib/Db/ShareWrapperRequest.php
```diff
@@ -431,12 +431,57 @@ public function delete(int $shareId): void {
...
+			$ids[] = $data['id'];
+		}
+		$cursor->closeCursor();
+
+		$this->deleteSharesAndChild($ids);
+	}
+
+
+	/**
+	 * @param array $ids
```
CarlSchwan, line 474:
```suggestion
	 * @param string[] $ids
```

if I read the code correctly this should be the correct type

ArtificialOwl: fixed

lib/Db/ShareWrapperRequest.php
```diff
@@ -431,12 +431,57 @@ public function delete(int $shareId): void {
...
+	 */
+	private function deleteSharesAndChild(array $ids): void {
+		$qb = $this->getShareDeleteSql();
+		$qb->andWhere(
+			$qb->expr()->orX(
+				$qb->exprLimitInArray('id', $ids),
```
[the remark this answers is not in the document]
ArtificialOwl: fixed, please review last commit !

CarlSchwan (approved):

PR #1006, opened 2022-04-11
split heavy and low maintenance process

ArtificialOwl:
fix #276

Circles already had a kind of level about maintenance, heavy process were already run during night.

This PR will create a new BackgroundJob for heavy maintenance processes (level 4 and 5) with TIME_INSENSITIVE

while the low maintenance processes will keep using TIME_SENSITIVE


lib/Cron/MaintenanceHeavy.php
```diff
@@ -0,0 +1,63 @@
...
+
+
+	/**
+	 * @param ITimeFactory $time
+	 * @param MaintenanceService $maintenanceService
+	 */
+	public function __construct(ITimeFactory $time, MaintenanceService $maintenanceService) {
+		parent::__construct($time);
+
+		$this->setInterval(10);
```
CarlSchwan, line 50: This needs to be every day, otherwise they are not really time insensitive ;)

ArtificialOwl: fixed. Also removed the night-time check before running heavy process, as it should be managed by the INSENSITIVE flag

CarlSchwan (approved):
Looks good to me :)

PR #1024, opened 2022-04-20
Add static analysis

CarlSchwan:
Currently, most of the issues reported by psalm are ignored, I just fixed a few one. But at least this will ensure the number of issues reported doesn't increase


lib/Circles/FileSharingBroadcaster.php
```diff
@@ -453,9 +453,8 @@ protected function sendMail($fileName, $link, $author, $circleName, $email) {
 		$senderName = $this->l10n->t('%s on %s', [$author, $instanceName]);
 
 		$message->setFrom([Util::getDefaultEmailAddress($instanceName) => $senderName]);
-		$message->setSubject($subject);
-		$message->setPlainBody($emailTemplate->renderText());
-		$message->setHtmlBody($emailTemplate->renderHtml());
+		$emailTemplate > setSubject($subject);
```
ArtificialOwl, line 456: welp ?

lib/Command/CirclesMaintenance.php
```diff
@@ -109,7 +109,7 @@ protected function execute(InputInterface $input, OutputInterface $output): int
 		$level = (int)$input->getOption('level');
 
 		if ($reset || $uninstall) {
-			$action = strtolower(($uninstall) ? 'uninstall' : 'reset');
+			$action = $uninstall ? 'uninstall' : 'reset';
```
ArtificialOwl, line 112: why not keeping the `strtolower()` ?

ArtificialOwl (approved):

lib/Command/CirclesMaintenance.php
CarlSchwan: because uninstall and reset are already lower case

lib/Circles/FileSharingBroadcaster.php
CarlSchwan: forgot the `-` in the `->` :(

lib/Command/CirclesMaintenance.php
ArtificialOwl: ok, so now I am wondering why I used `strtolower()` myself ...

PR #1025, opened 2022-04-21
fix array key on import()

ArtificialOwl:
`import()` is used to deserialize the model, array keys should fit the one used in `jsonSerialize()`


ArtificialOwl:
@CarlSchwan @nickvergessen quick review ?

CarlSchwan (approved):

PR #1052, opened 2022-06-13
Change return type to handle change from master

CarlSchwan:


ArtificialOwl (approved):

ArtificialOwl:
I don't think we need a backport on this one ?

CarlSchwan:
> I don't think we need a backport on this one ?

we don't

PR #1117, opened 2022-07-27
Add `$shallow` arg to `ShareByCircleProvider::getSharesInFolder`

artonge:
A new `$shallow` argument was introduced to `IShareProvider::getSharesInFolder` in NC25 by https://github.com/nextcloud/server/pull/31728.

This PR adds this new argument and the necessary changes to make it work based on this example https://github.com/nextcloud/server/pull/31728/files#diff-dee1351696d7eae959761edf66c8e5a9052e28be73319733d857a9ab2d9fc967=

Looks like in the context of `ShareWrapperRequest`, the `$shallow` argument can't be used as we only have the `nodeId` so we cannot use the same method as in other places:

```php
if ($shallow) {
	$qb->andWhere($qb->expr()->eq('f.parent', $qb->createNamedParameter($node->getId())));
} else {
	$qb->andWhere($qb->expr()->like('f.path', $qb->createNamedParameter($this->dbConnection->escapeLikeParameter($node->getInternalPath()) . '/%')));
}
```

Should we change the methods to forward `$node` instead of `$nodeId`?

### DIFF
--- a/lib/Db/ShareWrapperRequest.php
+++ b/lib/Db/ShareWrapperRequest.php
@@ -40,6 +40,7 @@ use OCA\Circles\Model\ShareWrapper;
 use OCP\Files\NotFoundException;
 use OCP\Share\Exceptions\IllegalIDChangeException;
 use OCP\Share\IShare;
+use OCP\Files\Folder;
 
 /**
  * Class ShareWrapperRequest
@@ -358,25 +359,30 @@ class ShareWrapperRequest extends ShareWrapperRequestBuilder {
 
 	/**
 	 * @param FederatedUser $federatedUser
-	 * @param int $nodeId
+	 * @param Folder $node
 	 * @param bool $reshares
+	 * @param bool $shallow Whether the method should stop at the first level, or look into sub-folders.
 	 *
 	 * @return ShareWrapper[]
 	 * @throws RequestBuilderException
 	 */
 	public function getSharesInFolder(
 		FederatedUser $federatedUser,
-		int $nodeId,
-		bool $reshares
+		Folder $node,
+		bool $reshares,
+		bool $shallow = true
 	): array {
 		$qb = $this->getShareSelectSql();
 
 		$qb->leftJoinCircle(CoreQueryBuilder::SHARE, null, 'share_with');
 		$qb->limitToShareOwner(CoreQueryBuilder::SHARE, $federatedUser, $reshares);
 		$qb->leftJoinFileCache(CoreQueryBuilder::SHARE);
-		if ($nodeId > 0) {
-			$aliasFileCache = $qb->generateAlias(CoreQueryBuilder::SHARE, CoreQueryBuilder::FILE_CACHE);
-			$qb->limitInt('parent', $nodeId, $aliasFileCache);
+
+		$aliasFileCache = $qb->generateAlias(CoreQueryBuilder::SHARE, CoreQueryBuilder::FILE_CACHE);
+		if ($shallow) {
+			$qb->limitInt('parent', $node->getId(), $aliasFileCache);
+		} else {
+			$qb->like('path', $node->getInternalPath() . '/%', $aliasFileCache);
 		}
 		$qb->limitNull('parent', false);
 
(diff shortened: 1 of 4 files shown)


lib/Db/ShareWrapperRequest.php
```diff
@@ -360,14 +360,16 @@ public function getSharesBy(
...
 	 *
 	 * @return ShareWrapper[]
 	 * @throws RequestBuilderException
 	 */
 	public function getSharesInFolder(
 		FederatedUser $federatedUser,
 		int $nodeId,
-		bool $reshares
+		bool $reshares,
+		$shallow = true
```
ArtificialOwl, line 372: bool ?

lib/Service/ShareWrapperService.php
```diff
@@ -269,12 +269,13 @@ public function getSharesBy(
...
 	 * @param int $nodeId
 	 * @param bool $reshares
+	 * @param bool $shallow Whether the method should stop at the first level, or look into sub-folders.
 	 *
 	 * @return ShareWrapper[]
 	 * @throws RequestBuilderException
 	 */
-	public function getSharesInFolder(FederatedUser $federatedUser, int $nodeId, bool $reshares): array {
-		return $this->shareWrapperRequest->getSharesInFolder($federatedUser, $nodeId, $reshares);
+	public function getSharesInFolder(FederatedUser $federatedUser, int $nodeId, bool $reshares, $shallow = true): array {
```
ArtificialOwl, line 277: bool ?

lib/ShareByCircleProvider.php
```diff
@@ -404,12 +404,13 @@ public function move(IShare $share, $recipient): IShare {
 	 * @throws RequestBuilderException
 	 * @throws SingleCircleNotFoundException
 	 */
-	public function getSharesInFolder($userId, Folder $node, $reshares): array {
+	public function getSharesInFolder($userId, Folder $node, $reshares, $shallow = true): array {
```
ArtificialOwl, line 407: bill ?

can you set variable type for each param ?

lib/ShareByCircleProviderDeprecated.php
```diff
@@ -378,10 +378,11 @@ private function createShareChild($userId, $share) {
 	 * @param Folder $node
 	 * @param bool $reshares Also get the shares where $user is the owner instead of just the
 	 *     shares where $user is the initiator
+	 * @param bool $shallow Whether the method should stop at the first level, or look into sub-folders.
 	 *
 	 * @return Share[]
 	 */
-	public function getSharesInFolder($userId, Folder $node, $reshares) {
+	public function getSharesInFolder($userId, Folder $node, $reshares, $shallow = true) {
```
ArtificialOwl, line 385: et bool !

ArtificialOwl:
Yes, it seems you can pass `Folder` (`Node` ?) instead of `int` in https://github.com/nextcloud/circles/blob/94b67c203a7a8ffaf2b277fe2fb9bec3cf7aebdc/lib/ShareByCircleProvider.php#L411 without affecting the rest of the app

ArtificialOwl:
also, for the sql requestion (around https://github.com/nextcloud/circles/blob/94b67c203a7a8ffaf2b277fe2fb9bec3cf7aebdc/lib/Db/ShareWrapperRequest.php#L382) you can test:


>     $qb->like('path', $node->getInternalPath() . '/%', $aliasFileCache);

lib/Db/ShareWrapperRequest.php
artonge: Yes, need a type if there is a default value ?

lib/ShareByCircleProvider.php
artonge: Ok

lib/Db/ShareWrapperRequest.php
ArtificialOwl: I don't know, but I personally like when it is set, even with a default value :)

lib/ShareByCircleProvider.php
artonge: Incompatible with the interface if I add bool, so I am leaving it like this, sorry :/

lib/ShareByCircleProviderDeprecated.php
artonge: Incompatible with the interface if I add bool, so I am leaving it like this, sorry :/

ArtificialOwl (approved):
no tested, looks good !

PR #1118, opened 2022-07-29
ignore no initiator

ArtificialOwl:


come-nc (approved):
When does a circle have no initiator?

ArtificialOwl:
excellent question, I think we are in a really specific use case with user not already existing locally, but being updated on ldap.
Because the method is only called when the the display name of the user is changed, it won't be a big issue if the display name is not updated if not available yet in database :]

PR #1088, opened 2022-06-18
+DataProbe

ArtificialOwl:
should fix #1039

This patch add an alternative to `CirclesManager::getCircles()` to retrieve data about Circles based on current session: `CirclesManager::probeCircles()`

- The access right is only based on Cached Memberships, making the `WHERE` condition of the request is a lot lighter,
- The method use `DataProbe` as parameter to add details to each circles returned by the request.

### DataProbe

it allows the app using `probeCircles()` to define exactly which data need to be returned.
The way `DataProbe` works is to edit the list of [`Sql Path` available](https://github.com/nextcloud/circles/blob/master/lib/Db/CoreQueryBuilder.php#L84) and only generate request based on this alternate list

>     $circles = $circlesManager->probeCircles();
```
[
  {
    "id": "k69QCDoZQFucbphq8Ha3JtNIdc4cR3g",
    "name": "oui2",
    "displayName": "oui2",
    "sanitizedName": "oui2",
    "source": 16,
    "population": 2,
    "config": 0,
    "description": "",
    "url": "/index.php/apps/contacts/direct/circle/k69QCDoZQFucbphq8Ha3JtNIdc4cR3g",
    "creation": 1654795649,
    "initiator": null
  }
]
```

>     $dataProbe = new DataProbe();
>     $dataProbe->add(DataProbe::OWNER);
>     $circles = $circlesManager->probeCircles(null, $dataProbe);
```
[
  {
    "id": "k69QCDoZQFucbphq8Ha3JtNIdc4cR3g",
    "name": "oui2",
    "displayName": "oui2",
    "sanitizedName": "oui2",
    "source": 16,
    "population": 2,
    "config": 0,
    "description": "",
    "url": "/index.php/apps/contacts/direct/circle/k69QCDoZQFucbphq8Ha3JtNIdc4cR3g",
    "creation": 1654795649,
    "initiator": null,
    "owner": {
      "id": "24csHL6SpJSp3C6x7CUkpZ1OVVgUqoF",
      "circleId": "k69QCDoZQFucbphq8Ha3JtNIdc4cR3g",
      "singleId": "RZXwUoc5jmTS5ECKuESJ7627uqIuioV",
      "userId": "test1",
      "userType": 1,
      "instance": "nc24.local",
      "local": true,
      "level": 9,
      "status": "Member",
      "displayName": "test1",
      "displayUpdate": 1654795649,
      "notes": {
        "invitedBy": {
[...]

PR #1155, opened 2022-09-09
cache getCircles when executed from an other app

ArtificialOwl:
this is a quick fix, before the implementation of DataProbe (#1088)


mejo- (approved):
Didn't test, but looks good codewise :+1:

PR #1169, opened 2022-09-30
refresh cache on demand

ArtificialOwl:
might not be the best solution

### DIFF
--- a/lib/CirclesManager.php
+++ b/lib/CirclesManager.php
@@ -322,14 +322,14 @@ class CirclesManager {
 	 * @throws InitiatorNotFoundException
 	 * @throws RequestBuilderException
 	 */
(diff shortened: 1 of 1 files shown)


max-nextcloud:
I think this would do the trick for collectives. However the better solution might be to expire the cache when a new circle is created.

We are using the `createCircle` function of the `circleManager` from collectives:
```php
$federatedUser = $this->getFederatedUser($userId);
$this->circlesManager->startSession($federatedUser);
$circle = $this->circlesManager->createCircle($name, null, false, false);
```

This should expire the cache for the `federatedUser` in question, right? Or should we use another method?

max-nextcloud:
Okay... let's go with this for a start.

PR #1188, opened 2022-10-20
update share on password change

ArtificialOwl:
- remove password on previous share if enforce password protection is disabled (globally, not just on current circle)
- update password on previous shares on static password changes (if password protection is enabled)
- adding '--test-password' option in ./occ circles:manage:setting to confirm password protection of shares
- notice log on getShareByToken if password protected, including hashed password

### DIFF
--- a/lib/ShareByCircleProvider.php
+++ b/lib/ShareByCircleProvider.php
@@ -85,6 +85,7 @@ use OCP\Share\Exceptions\IllegalIDChangeException;
 use OCP\Share\Exceptions\ShareNotFound;
 use OCP\Share\IShare;
 use OCP\Share\IShareProvider;
+use Psr\Log\LoggerInterface;
 
 /**
  * Class ShareByCircleProvider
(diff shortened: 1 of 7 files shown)


lib/ShareByCircleProvider.php
```diff
@@ -155,7 +124,7 @@ public function __construct(
 		$this->userManager = $userManager;
 		$this->rootFolder = $rootFolder;
 		$this->l10n = $l10n;
-		$this->logger = $logger;
+		$this->logger = OC::$server->get(LoggerInterface::class);
```
mejo-, line 127: Any reason to not use dependency injection here?

ArtificialOwl: legacy code on core :]

mejo-: Ok, but then you don't have to pass `$logger` into `__construct()`, do you?

mejo- (approved):
Looks good codewise appart from my comments.

lib/ShareByCircleProvider.php
ArtificialOwl: i don't want to touch legacy code right now, might fix for nc26

PR #1201, opened 2022-11-08
re-activate getMember()

ArtificialOwl:
reactivating getMember from old Circles' API to avoid conflict with current ShareAPIController from core


juliushaertl (approved):
Looks good and server seems to be the only place where this is being used according to https://sourcegraph.com/search?q=context:global+repo:github.com/nextcloud/*+::getMember&patternType=standard

PR #1202, opened 2022-11-09
remove shares on removing members

ArtificialOwl:
fixing an issue where shares were not removed when removing a member.

if a member is now removed, we checked for each inherited members if a share was done to the circle and parents.


mejo- (approved):
Apart from the lint error in CI it looks good to me :relaxed:

PR #1214, opened 2022-11-17
emulate entity on Admin OCS API

ArtificialOwl:
allow to emulate a federatedUser/entity when using the admin ocs api

### DIFF
--- a/lib/Controller/AdminController.php
+++ b/lib/Controller/AdminController.php
@@ -31,8 +31,6 @@ declare(strict_types=1);
 
 namespace OCA\Circles\Controller;
 
-use OCA\Circles\Tools\Traits\TDeserialize;
-use OCA\Circles\Tools\Traits\TNCLogger;
 use Exception;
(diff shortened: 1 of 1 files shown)


lib/Controller/AdminController.php
```diff
@@ -543,6 +541,18 @@ public function link(string $emulated, string $circleId, string $singleId): Data
 	private function setLocalFederatedUser(string $emulated): void {
 		$user = $this->userSession->getUser();
 		$this->federatedUserService->setCurrentPatron($user->getUID());
+
+		if (strpos($emulated, '=') > 0) {
+			[$type, $id] = explode('=', $emulated, 2);
+			switch (strtolower($type)) {
+				case 'singleid':
```
mejo-, line 548: Maybe something to document somewhere? Not sure whether there's documentation where it fits - but at least I'd document the logic in the annotations of this function, or in the OCS controller.

mejo- (changes requested):
Looks good to me codewise. But I think that documenting it somewhere is crucial.

PR #1221, opened 2022-11-25
[stable24] avoid exception loop

ArtificialOwl:


ArtificialOwl:
(psalm is not related to current ticket and comes from deprecated/unused code)

blizzz (approved):

PR #1224, opened 2022-11-30
check app is enabled

ArtificialOwl:
verify app set in route_to_circle is enabled

### DIFF
--- a/lib/Model/ModelManager.php
+++ b/lib/Model/ModelManager.php
@@ -31,7 +31,6 @@ declare(strict_types=1);
 
 namespace OCA\Circles\Model;
 
-use Exception;
 use OCA\Circles\AppInfo\Application;
 use OCA\Circles\Db\CircleRequest;
 use OCA\Circles\Db\CoreQueryBuilder;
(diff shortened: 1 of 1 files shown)


mejo- (approved):

lib/Model/ModelManager.php
```diff
@@ -554,16 +540,24 @@ public function fixInstance(string $instance): string {
...
+			$path = $this->configService->getAppValue(ConfigService::ROUTE_TO_CIRCLE);
+			$pos = strpos($path, '.');
+			if (!$pos) {
+				return '';
+			}
 
-		try {
-			if ($path !== '') {
-				return $this->urlGenerator->linkToRoute($path, ['singleId' => $singleId]);
+			if (!$this->appManager->isInstalled(substr($path, 0, $pos))) {
```
juliushaertl, line 551: This should check if the app is enabled, since routes are only loaded in that case.

juliushaertl (changes requested):

lib/Model/ModelManager.php
ArtificialOwl: isInstalled() check if app is enabled, as said in method phpdoc :)

juliushaertl: I'd say even better reason to use the other one to be more explicit 🙈 We should actually deprecate that app manager method then.

juliushaertl (approved):
Not blocking but I'd prefer the other method 👍

PR #1233, opened 2023-01-03
fix: Do not spam the log if contacts is disabled

juliushaertl:
Avoid spamming the log if the contacts app is disabled. I consider this more a workaround as I don't have enough insight on why the route is configurable in that case. Are there other use cases where the route could be served by another app? If not that could clean up the code quite a bit by hardcoding it to checking if contacts is enabled and then returning the route, otherwise none.


ArtificialOwl:
dupe of https://github.com/nextcloud/circles/pull/1224

juliushaertl:
Closing in favour of https://github.com/nextcloud/circles/pull/1224

PR #1241, opened 2023-01-26
Create API documentation from source code using phpDocumentor

susnux:
Currently there is no documentation so this provides at least some API documentation so users can search the `CirclesManager` interface without digging through the source code.


mejo- (approved):
Thanks a lot for the effort to start with documenting the PHP API of Circles @susnux :heart: 

I like the approach.

PR #1255, opened 2023-02-08
[stable23] check app is enabled

ArtificialOwl:
backport of #1224


blizzz:
@ArtificialOwl do we still need it?

PR #1269, opened 2023-02-24
exception on missing session

ArtificialOwl:


lib/CirclesManager.php
```diff
@@ -249,9 +249,15 @@ public function stopSession(): void {
 
 	/**
 	 * @return IFederatedUser
+	 * @throws FederatedUserNotFoundException
 	 */
 	public function getCurrentFederatedUser(): IFederatedUser {
-		return $this->federatedUserService->getCurrentUser();
+		$current = $this->federatedUserService->getCurrentUser();
+		if (is_null($current)) {
```
skjnldsv, line 256: Nitpick, `$current === null` is more efficient :)
We can merge as is, this is minimal

skjnldsv (approved):

PR #1291, opened 2023-04-25
refresh display name

ArtificialOwl:
maintenance level 5 should update display name of users if older than 8 days since last update. 
`--force-refresh` will bypass this 8 days check.

### DIFF
--- a/lib/Command/CirclesMaintenance.php
+++ b/lib/Command/CirclesMaintenance.php
@@ -91,7 +91,8 @@ class CirclesMaintenance extends Base {
 			 ->addOption(
 			 	'uninstall', '', InputOption::VALUE_NONE,
 			 	'Uninstall the apps and everything related to the app from the database'
-			 );
+			 )
+			->addOption('force-refresh', '', InputOption::VALUE_NONE, 'enforce some refresh');
 	}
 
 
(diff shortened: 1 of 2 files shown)


lib/Command/CirclesMaintenance.php
```diff
@@ -91,7 +91,8 @@ protected function configure() {
 			 ->addOption(
 			 	'uninstall', '', InputOption::VALUE_NONE,
 			 	'Uninstall the apps and everything related to the app from the database'
-			 );
+			 )
+			->addOption('force-refresh', '', InputOption::VALUE_NONE, 'enforce some refresh');
```
mejo-, line 95: Maybe it would be good to be more explicit what "some refresh" means?

lib/Service/MaintenanceService.php
```diff
@@ -217,7 +217,7 @@ private function lockMaintenanceRun(): void {
 	/**
 	 * every minute
 	 */
-	private function runMaintenance1(): void {
+	private function runMaintenance1(bool $forceRefresh = false): void {
```
mejo-, line 220: Is there a reason to pass `$forceRefresh` to all `runMaintenanceX()` functions even though it's only used in `runMaintenance5()`?

mejo- (approved):
Just two minor comments. Feel free to ignore.

PR #1294, opened 2023-05-04
app cfg from cli

ArtificialOwl:


mejo-:
Thanks, this indeed fixes setting the `CFG_APP` flag via `occ` command with `--super-session`. Without, it still doesn't work. Is this intended? I.e., is only super sessions allowed to set/unset `CFG_APP`?

ArtificialOwl:
yes, super session are needed to ensure it is not requested from a normal user but by an app

mejo- (approved):

PR #1347, opened 2023-07-18
generate path if app _DOES_ exist

ArtificialOwl:
condition was wrongly set: we want the app set as route to be enabled.


Altahrim (approved):

AndyScherzinger (approved):

AndyScherzinger:
@ArtificialOwl manual backports needed

PR #1349, opened 2023-07-19
Send notification mail for circle invitation

### LINKED ISSUE #235: Invite by notifications and email
It is necessary that invitations to the circle arrive by "nextcloud notifications" and by email.

onny:
Get notified if someone invites you into a circle.

image

Fixes https://github.com/nextcloud/circles/issues/235

### DIFF
--- a/lib/Listeners/Notifications/RequestingMember.php
+++ b/lib/Listeners/Notifications/RequestingMember.php
@@ -37,8 +37,10 @@ use OCA\Circles\Events\CircleGenericEvent;
 use OCA\Circles\Events\RequestingCircleMemberEvent;
 use OCA\Circles\Exceptions\RequestBuilderException;
 use OCA\Circles\Service\NotificationService;
+use OCA\Circles\Service\SendMailService;
 use OCP\EventDispatcher\Event;
 use OCP\EventDispatcher\IEventListener;
+use OCP\IUserManager;
 
 /**
  * Class RequestingMember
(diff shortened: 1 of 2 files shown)


onny:
Will be happy for a review, it's my first contribution to Circles @ArtificialOwl 😊

onny:
Happy for review @susnux :)

lib/Service/SendMailService.php
```diff
@@ -88,6 +88,93 @@ public function __construct(
...
+			$author,
+			$circleName
+		);
+
+		foreach ($mails as $mail) {
+			try {
+				$this->sendMailInvitation($template, $author, $mail, $circleName);
+			} catch (Exception $e) {
+			}
+		}
```
Fenn-CS, line 125: What happens during this exception? You probably want to write this on the logs?

lib/Service/SendMailService.php
```diff
@@ -88,6 +88,93 @@ public function __construct(
...
+
+		$message = $this->mailer->createMessage();
+
+		$message->setFrom([Util::getDefaultEmailAddress($instanceName) => $senderName]);
+		$message->setSubject($subject);
+		$message->setPlainBody($emailTemplate->renderText());
+		$message->setHtmlBody($emailTemplate->renderHtml());
+		$message->setTo([$recipient]);
+
+		$this->mailer->send($message);
```
Fenn-CS, line 174: I don't know but it could be possible to build all the mail templates and call the send function with an array of messages? (Related to the loop that calls `sendMailInvitation` on linr 122)

nfebe (changes requested):
A few inline comments.

lib/Listeners/Notifications/RequestingMember.php
```diff
@@ -74,11 +87,28 @@ public function handle(Event $event): void {
...
 		if ($event->getType() === CircleGenericEvent::REQUESTED) {
 			$this->notificationService->notificationRequested($member);
 		} else {
 			$this->notificationService->notificationInvited($member);
+
+			if ($member->hasInvitedBy()) {
+				$author = $member->getInvitedBy()->getDisplayName();
+			} else {
+				$author = 'someone';
+			}
```
Fenn-CS, line 101:
```suggestion
			}
			$author = ucfirst($author);
```

I see in the screen shot that `test` started with lower-case so, we want to start the statement and name in this case with caps.

susnux:
Not sure if this is the best way to implement this, currently we using notifications, but activities would support mail out of the box (users can dis / enable mails). So we would not need to implement any mail logic here.

@ArtificialOwl you implemented the activities already but commented out it in the app info xml, should we try to enable them again?

ArtificialOwl:
> @ArtificialOwl you implemented the activities already but commented out it in the app info xml, should we try to enable them again?

I am currently working on it

PR #1357, opened 2023-08-03
download permissions on share for internal/external members

ArtificialOwl:
manage `allow download` on shares to Circles using the same mechanism that internal shares and shareByMail.

- local account as members of a Circle will not be able to download the shared content,
- remote account (mail address) as members will not be able to see the download button.

the attributes is stored/converted the same way than in core.

### DIFF
--- a/lib/Db/ShareWrapperRequest.php
+++ b/lib/Db/ShareWrapperRequest.php
@@ -31,16 +31,18 @@ declare(strict_types=1);
 
 namespace OCA\Circles\Db;
 
+use JsonException;
 use OCA\Circles\Exceptions\RequestBuilderException;
 use OCA\Circles\Exceptions\ShareWrapperNotFoundException;
 use OCA\Circles\Model\FederatedUser;
 use OCA\Circles\Model\Membership;
 use OCA\Circles\Model\Probes\CircleProbe;
 use OCA\Circles\Model\ShareWrapper;
+use OCP\Files\Folder;
 use OCP\Files\NotFoundException;
 use OCP\Share\Exceptions\IllegalIDChangeException;
+use OCP\Share\IAttributes;
 use OCP\Share\IShare;
-use OCP\Files\Folder;
 
 /**
  * Class ShareWrapperRequest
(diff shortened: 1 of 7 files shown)


lib/Db/ShareWrapperRequest.php
```diff
@@ -95,12 +96,16 @@ public function save(IShare $share, int $parentId = 0): int {
...
 		$qb = $this->getShareUpdateSql();
+		$shareAttributes = $this->formatShareAttributes($shareWrapper->getAttributes());
+
 		$qb->set('file_target', $qb->createNamedParameter($shareWrapper->getFileTarget()))
 		   ->set('share_with', $qb->createNamedParameter($shareWrapper->getSharedWith()))
 		   ->set('uid_owner', $qb->createNamedParameter($shareWrapper->getShareOwner()))
 		   ->set('uid_initiator', $qb->createNamedParameter($shareWrapper->getSharedBy()))
 		   ->set('accepted', $qb->createNamedParameter(IShare::STATUS_ACCEPTED))
-		   ->set('permissions', $qb->createNamedParameter($shareWrapper->getPermissions()));
+		   ->set('permissions', $qb->createNamedParameter($shareWrapper->getPermissions()))
```
Altahrim, line 106: Nit: small indent problem

lib/Db/ShareWrapperRequest.php
```diff
@@ -486,4 +491,25 @@ private function deleteSharesAndChild(array $ids): void {
...
 		$qb->execute();
 	}
+
+
+	/**
+	 * Format IAttributes to database format (JSON string)
+	 * based on OC\Share20\DefaultShareProvider::formatShareAttributes();
+	 */
+	private function formatShareAttributes(?IAttributes $attributes): ?string {
+		if ($attributes === null || empty($attributes->toArray())) {
```
Altahrim, line 501: Nit: 
```suggestion
		if (empty($attributes?->toArray())) {
```

lib/Db/ShareWrapperRequest.php
```diff
@@ -486,4 +491,25 @@ private function deleteSharesAndChild(array $ids): void {
...
+			return null;
+		}
+
+		$compressedAttributes = [];
+		foreach ($attributes->toArray() as $attribute) {
+			$compressedAttributes[] = [
+				0 => $attribute['scope'],
+				1 => $attribute['key'],
+				2 => $attribute['enabled']
```
Altahrim, line 510: Is there a reason to force indexes?
Could be: 
```php
		$compressedAttributes[] = [
				$attribute['scope'],
				$attribute['key'],
				$attribute['enabled']
			];

File: lib/Db/ShareWrapperRequest.php [hunk 4]
```diff
@@ -486,4 +491,25 @@ private function deleteSharesAndChild(array $ids): void {
...
+			];
+		}
+		return \json_encode($compressedAttributes);
```
Altahrim, line 513: Nit: 
```suggestion
		return \json_encode($compressedAttributes,  JSON_UNESCAPED_SLASHES|JSON_UNESCAPED_UNICODE|JSON_THROW_ON_ERROR);
1```

File: lib/Model/ShareWrapper.php [hunk 5]
```diff
@@ -541,12 +570,43 @@ public function importFromDatabase(array $data, string $prefix = ''): IQueryRow
...
+		}
+
+		$attributes = new ShareAttributes();
+		$compressedAttributes = json_decode($data, true);
+		if (!is_array($compressedAttributes)) {
+			return;
+		}
+
+		foreach ($compressedAttributes as $compressedAttribute) {
+			$attributes->setAttribute(
```
Altahrim, line 590: Shorter version: 
```suggestion
			$attributes->setAttribute(...$compressedAttribute);
```
A bit different if other arguments are available…

Altahrim (approved):

PR #1369, opened 2023-08-19
support sharing files_external root

ArtificialOwl:
This fix an issue when sharing the top folder of an external storage.
path is empty in that case, meaning we should not return `false` when checking availability if path is empty but storage is managed by `files_external`

### DIFF
--- a/lib/Model/FileCacheWrapper.php
+++ b/lib/Model/FileCacheWrapper.php
@@ -31,13 +31,13 @@ declare(strict_types=1);
 
 namespace OCA\Circles\Model;
 
+use JsonSerializable;
+use OCA\Circles\Db\CoreQueryBuilder;
+use OCA\Circles\Exceptions\FileCacheNotFoundException;
 use OCA\Circles\Tools\Db\IQueryRow;
 use OCA\Circles\Tools\Exceptions\InvalidItemException;
 use OCA\Circles\Tools\IDeserializable;
 use OCA\Circles\Tools\Traits\TArrayTools;
-use JsonSerializable;
-use OCA\Circles\Db\CoreQueryBuilder;
-use OCA\Circles\Exceptions\FileCacheNotFoundException;
 
 /**
  * Class FileCacheWrapper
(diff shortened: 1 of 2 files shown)


lib/Model/FileCacheWrapper.php
```diff
@@ -451,12 +451,19 @@ public function toCache(): array {
...
 		}
 
-		return !(explode('/', $this->getPath(), 2)[0] !== 'files'
-				 && explode(':', $this->getStorage(), 2)[0] === 'home');
+		$path = $this->getPath();
+		[$storageType,] = explode('::', $this->getStorage(), 2);
+
+		if ($path === '') {
+			// we only accept empty path on external storage
+			return (in_array($storageType, ['ftp', 'sftp', 'swift', 'smb', 'amazon']));
```
Altahrim, line 463: Do we have this list as a dynamic array somewhere?
I am afraid we will forget to update this one if we support a new external storage

Altahrim (approved):

lib/Model/FileCacheWrapper.php
ArtificialOwl: @Altahrim added a notice in case of an eventual unsupported storage
@szaimen  I don't think you can share your full (top-root) home

szaimen: What I meant was that local storage is also a kind of external storage :)

szaimen:
```suggestion
			return (in_array($storageType, ['ftp', 'sftp', 'swift', 'smb', 'amazon', 'local']));
```

szaimen: btw, arent there even further kinds?

ArtificialOwl: @szaimen ok, got it, fixing :D

ArtificialOwl: added local and webdav

Fenn-CS (approved):

PR #1378, opened 2023-08-30
circles' activity

ArtificialOwl:
_Is compatible with globalscale/multi-instance setup_

generate Activity on following action:

- if `circles creation_activity` is set, when creating a new circle, to everyone
- when a circle is destroyed, to all its members
- when a local account is added to a circle, to all (inherited) moderators and the added local account
- when a local account join a circle, to all (inherited) moderators and the joining local account
- when a circle is added to a circle, to all (inherited) moderators and to all (inherited) members of the added circle
- when a circle join another circle, to all (inherited) moderators and to all (inherited) members of the joining circle
- when a member level is edited, to all (inherited) moderators and the affected member
- when ownership is transferred, to all (inherited) users
- when a member is removed from a circle, to all (inherited) moderators and the removed member


lib/Service/ActivityService.php
```diff
@@ -0,0 +1,375 @@
...
+	 * - if user is IUser, we get userId,
+	 * - if user is Member, we ignore non-local account and returns local userId,
+	 * - others models are ignored
+	 * - duplicate are ignored
+	 *
+	 * @param IEvent $event
+	 * @param array<IUser|IFederatedUser> $users
+	 */
+	private function publishEvent(IEvent $event, array $users): void {
+		$knownSingleIds = [];
```
ArtificialOwl, line 354: more comments

PR #1399, opened 2023-09-19
[stable25] update single display-name

ArtificialOwl:


blizzz:
do we still need it? Needs a review.

PR #1410, opened 2023-10-11
order by viewer level

ArtificialOwl:


sorbaugh (approved):

skjnldsv (approved):

lib/Db/CircleRequest.php
```diff
@@ -184,6 +184,7 @@ public function getCircles(?IFederatedUser $initiator, CircleProbe $probe): arra
 		$qb->filterCircles(CoreQueryBuilder::CIRCLE, $probe);
 		if (!is_null($initiator)) {
 			$qb->limitToInitiator(CoreQueryBuilder::CIRCLE, $initiator);
+			$qb->orderBy($qb->generateAlias(CoreQueryBuilder::CIRCLE, CoreQueryBuilder::INITIATOR) . '.level', 'desc');
```
skjnldsv, line 187: This doesn't seems to work for the front, I still have to sort https://github.com/nextcloud/contacts/blob/19f836ee23f726b919134c6642af5b8983df871e/src/components/AppNavigation/RootNavigation.vue#L351-L356

ArtificialOwl: Do you have cache ? can you try (with the right userId) :
>    ./occ circles:manage:list --initiator-type user --initiator <userId>

This should returns the same list than the ocs route GET /circles

skjnldsv: | Back | Front |
|:---------:|:------:|
|image|image|


https://github.com/nextcloud/contacts/pull/3665

skjnldsv:
@ArtificialOwl I thought that wasn't the proper fix? Why is this merged?

ArtificialOwl:
@skjnldsv i could not reproduce your issue, i investigate any eventual rewriting of orderBy and could not find any reason for it to fail on your computer. Every scenario I have tested were good. Do you still have the issue ? can you share me your database; at least the table `oc_circles_*` ?

skjnldsv:
> Every scenario I have tested were good

Ok then! :+1:

PR #1415, opened 2023-10-16
escape html

ArtificialOwl:


sorbaugh (approved):

js/files/circles.files.list.js
```diff
@@ -124,11 +124,20 @@
...
+						return this.escapeHTML(circle.name);
+					},
+
+					escapeHTML: function (text) {
+						return text.toString()
+							.split('&').join('&amp;')
+							.split('<').join('&lt;')
+							.split('>').join('&gt;')
+							.split('"').join('&quot;')
+							.split('\'').join('&#039;');
```
Fenn-CS, line 140:
```suggestion
  						const parser = new DOMParser();
  						const doc = parser.parseFromString(text, 'text/html');
  						return doc.body.textContent;
```

Fenn-CS: This would escape any chars not thought about and handle more edge cases, it however looks okay

Fenn-CS (approved):
Add a similar function is shorter and would handle all edge cases for your testing and consideration but his looks good too! Thanks

PR #1481, opened 2024-01-16
remove useless javascript

ArtificialOwl:
since 28, shared to circles should be implemented in unified search


joshtrichards:
Conflicts

PR #1483, opened 2024-01-17
fix search on circles share

ArtificialOwl:
this will add Circles entries to the shares list in the right panel


artonge (approved):
What does `$nodeId === 0`?
I feel like this is the same as before, but with an extra condition on `$nodeId === 0`

ArtificialOwl:
> What does `$nodeId === 0`? I feel like this is the same as before, but with an extra condition on `$nodeId === 0`

get global details, not specific to a file

PR #1520, opened 2024-03-05
fix(dashboard): Add `icon-teams` CSS class and fix path to app icon

mejo-:
Fixing the path to the app icon fixes dashboard widgets. Otherwise getting the widgets throws an error and all widgets are broken.

Adding the `icon-teams` CSS class fixes displaying the teams icon in the teams dashboard widget.

2024-03-05T17:36:08,669081871+01:00

### DIFF
--- a/lib/Dashboard/TeamDashboardWidget.php
+++ b/lib/Dashboard/TeamDashboardWidget.php
@@ -147,7 +147,7 @@ class TeamDashboardWidget implements IAPIWidgetV2, IIconWidget, IButtonWidget, I
 	}
 
 	public function getIconUrl(): string {
-		return $this->urlGenerator->getAbsoluteURL($this->urlGenerator->imagePath('circles', 'app.svg'));
+		return $this->urlGenerator->getAbsoluteURL($this->urlGenerator->imagePath('circles', 'circles.svg'));
 	}
 
 	private function getTeamPage(): string {


lib/Events/BeforeTemplateRenderedListener.php
```diff
@@ -0,0 +1,20 @@
...
+
+namespace OCA\Circles\Events;
+
+use OCA\Circles\AppInfo\Application;
+use OCP\AppFramework\Http\Events\BeforeTemplateRenderedEvent;
+use OCP\EventDispatcher\Event;
+use OCP\EventDispatcher\IEventListener;
+use OCP\Util;
+
+class BeforeTemplateRenderedListener implements IEventListener {
```
juliushaertl, line 13:
```suggestion
/** @template-implements IEventListener<BeforeTemplateRenderedEvent> */
class BeforeTemplateRenderedListener implements IEventListener {
```

lib/Dashboard/TeamDashboardWidget.php
```diff
@@ -147,7 +147,7 @@ public function getWidgetButtons(string $userId): array {
 	}
 
 	public function getIconUrl(): string {
-		return $this->urlGenerator->getAbsoluteURL($this->urlGenerator->imagePath('circles', 'app.svg'));
+		return $this->urlGenerator->getAbsoluteURL($this->urlGenerator->imagePath('circles', 'circles.svg'));
```
juliushaertl, line 150: Thanks, I somehow had a file with that name locally but not checked in

juliushaertl:
Just looked into moving over files_sharing, which would actually also have the class defined now in core: https://github.com/nextcloud/server/pull/44007 maybe we can then skip loading an additional css file

lib/Events/BeforeTemplateRenderedListener.php
```diff
@@ -0,0 +1,20 @@
...
+	public function handle(Event $event): void {
+		if (!($event instanceof BeforeTemplateRenderedEvent)) {
+			return;
+		}
+		Util::addStyle(Application::APP_ID, 'teams');
```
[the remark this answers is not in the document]
mejo-: Indeed. But I agree that using the one from server is better, so I removed the logic to load the extra CSS file.

skjnldsv (approved):

Issue #1434, opened 2023-11-07
Javascript error on 28


juliusknorr:
Likely related to the files 2 vue migration:

```
Uncaught TypeError: OCA.Files.FileList is undefined
    <anonymous> https://nextcloud.local/apps/circles/js/files/circles.files.list.js?v=7916aede-0:42
    <anonymous> https://nextcloud.local/apps/circles/js/files/circles.files.list.js?v=7916aede-0:306
[circles.files.list.js:42:36](https://nextcloud.local/apps/circles/js/files/circles.files.list.js?v=7916aede-0)
    <anonymous> https://nextcloud.local/apps/circles/js/files/circles.files.list.js?v=7916aede-0:42
    <anonymous> https://nextcloud.local/apps/circles/js/files/circles.files.list.js?v=7916aede-0:306
```

cc @ArtificialOwl

skjnldsv:
Yep, needs an update for F2V compatibility

benklaasen:
Does this break access to files and dirs marked as favourites prior to the upgrade to 28?

mwinkens:
This also breaks access to selected files

VGerris:
I see this error in 28.0.3 of Nextcloud after a recent update. It breaks UI things for file copy and move it seems.
Other errors encountered are this one https://github.com/nextcloud/server/issues/43953 which seems to be merged but not in 28.0.3 - perhaps that is related ? 
I also see Error: Invalid owner type when using right click - Move or Copy which points to a variety of closed bugs.

Also : https://docs.nextcloud.com/server/latest/developer_manual/app_publishing_maintenance/app_upgrade_guide/upgrade_to_28.html#removed-apis mentions :
OCA.Files: Everything removed but Sidebar and Settings. See the Added API section for replacements.

There is no mention on how to replace the specific OCA.Files.FileList in that document as far as I can see.

skjnldsv:
Still happening on 30 @ArtificialOwl 

```
TypeError: OCA.Files.FileList is undefined
    <anonymous> https://domain.com/apps/circles/js/files/circles.files.list.js:42
    <anonymous> https://domain.com/apps/circles/js/files/circles.files.list.js:306
```

skjnldsv:
But fixed on 30 and above it seems

PR #1521, opened 2024-03-06
fix/remove broken scripts

juliushaertl:
We need to migrate those scripts to the new files API but this is more work there is currently no time. It seems those scripts can cause failures as reported in https://github.com/nextcloud/circles/issues/1434

As they don't work right now there is no need to load them and this also has been a regular annoyance when running cypress tests locally where the random console errors cause failures.

Left the script files in place to know what to implement with the new API.

### DIFF
--- a/lib/Listeners/Files/ListenerFilesLoadScripts.php
+++ b/lib/Listeners/Files/ListenerFilesLoadScripts.php
@@ -45,8 +45,9 @@ class ListenerFilesLoadScripts implements IEventListener {
 			return;
 		}
 
-		Util::addScript('circles', 'files/circles.files.app');
-		Util::addScript('circles', 'files/circles.files.list');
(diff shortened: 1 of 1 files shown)


VGerris:
Just some comments, thanks for linking to the other issue. 
What do you mean with no time ? As in lower priority?
If anyone has a good quick getting-starting to developing that part, I might give it a shot.
Perhaps good to rename those scripts with something in the name indicating they need to be reimplemented or removed , like filename.a -> filename-disabled-needs-api-impl-ticket1521.a or something.
And also good to determine impact of this, it may be quite serious and as such be more priority than initially thought.
Thank you!

mejo- (approved):
Thanks @juliushaertl, I had planned to do something similar :see_no_evil:

PR #1524, opened 2024-03-06
[stable26] circles' activity

ArtificialOwl:
backport of #1378


emoral435 (approved):
Haven't tested but the changes are the same, so LGTM

PR #1525, opened 2024-03-07
feat: Implement team resource provider for shared files

juliushaertl:
Example screenshot from the team overview in contacts:

Screenshot 2024-03-07 at 17 16 48


mejo- (approved):
Code changes look good to me. Only thing I don't understand is why `CirclesManager` could be `null`.

juliushaertl:
Ah that was because I started the class in the files_sharing app but thought it fits better in circles as then its just disabled if the app is.

PR #1544, opened 2024-04-09
fix setIcon using svg

ArtificialOwl:


lib/Activity/Provider.php
```diff
@@ -93,10 +93,7 @@ private function initActivityParser(IEvent $event, array $params): void {
 	 * @param IEvent $event
 	 */
 	private function setIcon(IEvent $event): void {
-		$path = $this->urlGenerator->imagePath(
-			Application::APP_ID,
-			'circles.' . $this->activityManager->getRequirePNG() ? 'png' : 'svg'
```
artonge, line 98: Why is the check not needed anymore?

ArtificialOwl: we have no more png anymore since 22, and it seems to be fine having svg for the generated event

artonge (approved):

mejo-:
stable29 and stable28 don't need backports?

blizzz:
> stable29 and stable28 don't need backports?

This went against stable28, newer versions are already adjusted.

PR #1620, opened 2024-07-01
[stable27] fix setIcon using svg

blizzz:
backport of #1544


nickvergessen:
The iOS app was unable to show SVGs in the past.
But it seems they don't show any icons anymore apart from add/remove of a file.

mejo- (approved):

PR #1641, opened 2024-07-26
feat: dismiss notification when accepting an invitation

kesselb:
For https://github.com/nextcloud/circles/issues/810

I only tested the case (circle needs invitation, user is invited, user press accept, notification is gone). 
Needs a careful review.

PR #1650, opened 2024-07-31
fix(notifications): Don't trigger invite notifications for system groups

nickvergessen:
Regression from #1641

### Before
Adding a user to a system group triggers a notification:
> You have been invited by Circles into the Circle "group1"

### After

:broom:


kesselb (approved):
Thanks :+1: 

Sorry, I should have used an own event listener right away to avoid conflicts.

PR #1671, opened 2024-08-14
feat(deps): Add Nextcloud 31 support on master

nickvergessen:
Signed-off-by: Joas Schilling <coding@schilljs.com>

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -12,7 +12,7 @@
 Those groups of people can then be used by any other app for sharing purpose.
 		]]>
 	</description>
-	<version>30.0.0-dev</version>
+	<version>31.0.0-dev.0</version>
 	<licence>agpl</licence>
 	<author>Maxence Lange</author>
 	<types>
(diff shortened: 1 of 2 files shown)


appinfo/info.xml
```diff
@@ -12,7 +12,7 @@
 Those groups of people can then be used by any other app for sharing purpose.
 		]]>
 	</description>
-	<version>30.0.0-dev</version>
+	<version>31.0.0-dev.0</version>
```
mejo-, line 15: Maybe we should correct this to `31.0.0` @ArtificialOwl? Given that `-dev` usually is not removed before the final releases, I think it rather causes more confusion than it helps. What do you think?

mejo- (changes requested):

appinfo/info.xml
nickvergessen: Problem is one can not "go back", so if a migration would be added the version would be "dirty" on release.

Feel free to do in a follow up, I'll merge it anyway to fix daily and our other master/main deployed instances

PR #1674, opened 2024-08-19
refactor: Migrate away from deprecated `ILogger` to PSR-3

susnux:
Mostly replace `ILogger` with `LoggerInterface` and some minor cleanup (constructor property promotion).

Some places used the deprecated `logException`, this is easy to migrate: Simply use the appropriate log level on the logger and place the exception under the `exception` key in the context.

### DIFF
--- a/lib/ShareByCircleProvider.php
+++ b/lib/ShareByCircleProvider.php
@@ -55,7 +55,6 @@ use OCP\Files\Node;
 use OCP\Files\NotFoundException;
 use OCP\IDBConnection;
 use OCP\IL10N;
-use OCP\ILogger;
 use OCP\IURLGenerator;
 use OCP\IUserManager;
 use OCP\Security\ISecureRandom;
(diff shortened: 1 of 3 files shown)


lib/ShareByCircleProvider.php
```diff
@@ -98,7 +97,7 @@ public function __construct(
 		IUserManager $userManager,
 		IRootFolder $rootFolder,
 		IL10N $l10n,
-		ILogger $logger,
+		mixed $logger, // unused, only kept for compatibility with server
```
come-nc, line 100: Damn this is ugly.
Server should use DI and the app should register the service if needed.

come-nc (approved):

lib/ShareByCircleProvider.php
susnux: Yes I know but there are also other parameters where this is already done :pensive:

susnux: But server already checks `class_exists`, meaning it can be auto-loaded, so we should theoretically be able to simply DI in server.

PR #1731, opened 2024-10-29
fix(ShareWrapperRequest): Save share attributes on creation

nfebe:
- Before this commit, attributes added to a share were ignored when the share type was team/circles (share type `7`).

- Share attributes are not expected to contain any key or property named `enabled`.

Partial resolution of: https://github.com/nextcloud/server/issues/46920

This is a partial resolution because setting the `download` attribute may not fully enforce the correct download permissions. A follow-up PR in the server may be needed to update the `hideDownload` property of the share based on the value of `download` in the share attributes.

### DIFF
--- a/lib/Model/ShareWrapper.php
+++ b/lib/Model/ShareWrapper.php
@@ -478,6 +478,8 @@ class ShareWrapper extends ManagedModel implements IDeserializable, IQueryRow, J
 			->setToken($this->get('token', $data))
 			->setShareTime($shareTime);
 
+		$this->importAttributesFromDatabase($this->get('attributes', $data));
+
 		try {
 			$this->setExpirationDate(new DateTime($this->get('expiration', $data)));
 		} catch (\Exception $e) {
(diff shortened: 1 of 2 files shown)


artonge (approved):

lib/Model/ShareWrapper.php
```diff
@@ -590,7 +592,7 @@ public function jsonSerialize(): array {
 			'shareType' => $this->getShareType(),
 			'providerId' => $this->getProviderId(),
 			'permissions' => $this->getPermissions(),
-			'attributes' => $this->getAttributes(),
+			'attributes' => json_encode($this->getAttributes()->toArray()),
```
nickvergessen, line 595: This broke production:
```
Call to a member function toArray() on null
```

nickvergessen:
```suggestion
			'attributes' => $this->getAttributes() !== null ? json_encode($this->getAttributes()->toArray()) : null,
```

PR #1736, opened 2024-10-31
fix toArray on null

ArtificialOwl:
fix https://github.com/nextcloud/circles/pull/1731/files#r1825161374


nfebe (approved):

nfebe (commented):
Have a minor improvement to add here.

nfebe:
Resolves : https://github.com/nextcloud/circles/pull/1737

nfebe:
> Have a minor improvement to add here.

Discarded as `getAttributes` returns either an object or null, other type checks which is what I had in mind are unnecessary.

Issue #1735, opened 2024-10-31
Call to a member function toArray() on null


nickvergessen:
This broke production:
```
Call to a member function toArray() on null
```

_Originally posted by @nickvergessen in https://github.com/nextcloud/circles/pull/1731#discussion_r1825161374_

nickvergessen:
Temporary patch I applied:
https://github.com/nextcloud/circles/pull/1731#discussion_r1825163536

nickvergessen:
- Fixed by #1736

PR #1737, opened 2024-10-31
fix(ShareWrapper): Prevent attempt to access property on null

nfebe:
`IShare:getAttributes()` can return null

Resolves: https://github.com/nextcloud/circles/issues/1735


nfebe:
Closing in favor of : https://github.com/nextcloud/circles/pull/1736

PR #1775, opened 2024-11-26
new daily maintenance to delete old federated events

ArtificialOwl:
keep table clean: delete 1 month old federated event that are noted as terminated


mejo- (approved):
Looks good to me!

PR #1794, opened 2024-12-02
catch undef recipient 

ArtificialOwl:


mejo- (approved):
Seems to be cosmetic (or at least affecting the output) only, so approving without exactly understanding when `$recipient->getBasedOn()` would be falsy.

ArtificialOwl:
when a user is deleted, before the maintenance cronjob is executed to remove orphan share

PR #1809, opened 2024-12-11
fix: do not override sharing text

SebastianKrupinski:
Determined sharing text from if block above is overridden.


ChristophWurst (approved):
Makes sense

PR #1840, opened 2025-01-14
fix: add cache to joinedCircles API request

tcitworld:
This deprecated API is only used by DAV, but doesn't benefit from the cache, which makes it VERY heavy when you have a lot of users using calendar intensively


mejo- (approved):
To me this looks sensible. @ArtificialOwl what do you think?

ArtificialOwl:
Hello,

this has been fixed using `probeCircles` instead of `getCircles` which is more efficient.
Caching will also be implemented to `probeCircles`.

Issue #1714, opened 2024-10-18
OCA\\Circles\\Model\\ShareWrapper::getCircle(): Return value must be of type OCA\\Circles\\Model\\Circle, null returned in file 'apps/circles/lib/Model/ShareWrapper.php' line 272


kesselb:
```json
{
  "reqId": "ZwZErpX2sEUWuAB-zRuerAAAAEE",
  "level": 3,
  "time": "2024-10-09T10:54:06+02:00",
  "remoteAddr": "",
  "user": "--",
  "app": "index",
  "method": "POST",
  "url": "/index.php/apps/richdocuments/wopi/files/248046682_oc9efcj9n0v6/contents?access_token=&access_token_ttl=0",
  "message": "Exception thrown: Exception",
  "userAgent": "COOLWSD HTTP Agent 23.05.15.1",
  "version": "27.1.11.8",
  "exception": {
    "Exception": "Exception",
    "Message": "OCA\\Circles\\Model\\ShareWrapper::getCircle(): Return value must be of type OCA\\Circles\\Model\\Circle, null returned in file 'apps/circles/lib/Model/ShareWrapper.php' line 272",
    "Code": 0,
    "Trace": [
      {
        "file": "lib/private/AppFramework/App.php",
        "line": 183,
        "function": "dispatch",
        "class": "OC\\AppFramework\\Http\\Dispatcher",
        "type": "->"
      },
      {
        "file": "lib/private/Route/Router.php",
        "line": 315,
        "function": "main",
        "class": "OC\\AppFramework\\App",
        "type": "::"
      },
      {
        "file": "lib/base.php",
        "line": 1068,
        "function": "match",
        "class": "OC\\Route\\Router",
        "type": "->"
      },
      {
        "file": "index.php",
        "line": 38,
        "function": "handleRequest",
        "class": "OC",
        "type": "::"
      }
    ],
    "File": "lib/private/AppFramework/Http/Dispatcher.php",
    "Line": 169,
    "Previous": {
      "Exception": "TypeError",
      "Message": "OCA\\Circles\\Model\\ShareWrapper::getCircle(): Return value must be of type OCA\\Circles\\Model\\Circle, null returned",
      "Code": 0,
      "Trace": [
        {
          "file": "apps/circles/lib/ShareByCircleProvider.php",
          "line": 694,
          "function": "getCircle",
          "class": "OCA\\Circles\\Model\\ShareWrapper",
          "type": "->"
        },
        {
[...]

PR #1862, opened 2025-02-11
fix(occ): handle share wrappers that do not have a circle

max-nextcloud:
See https://github.com/nextcloud/circles/issues/1714#issuecomment-2602064835

### DIFF
--- a/lib/Command/SharesFiles.php
+++ b/lib/Command/SharesFiles.php
@@ -203,15 +203,20 @@ class SharesFiles extends Base {
 					$sharedTo .= '@' . $recipient->getInstance();
 				}
 			}
-			$circle = $share->getCircle();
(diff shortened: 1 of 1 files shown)


lib/Command/SharesFiles.php
```diff
@@ -203,15 +203,20 @@ private function displayShares(
 					$sharedTo .= '@' . $recipient->getInstance();
 				}
 			}
-			$circle = $share->getCircle();
+			$circleField = '';
+			if ($share->hasCircle()) {
```
max-nextcloud, line 207: I don't quite understand when a `ShareWrapper` has a circle and when it does not. We might also want to skip all `$share`s without a circle.

juliusknorr (approved):

PR #1866, opened 2025-02-12
fix: Do not send a mail if public share is disabled

artonge:
This prevent confusing users with emails containing link that lead to an error page.

### DIFF
--- a/lib/Service/SendMailService.php
+++ b/lib/Service/SendMailService.php
@@ -22,6 +22,7 @@ use OCP\IL10N;
 use OCP\Mail\IEMailTemplate;
 use OCP\Mail\IMailer;
 use OCP\Security\IHasher;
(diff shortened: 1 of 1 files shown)


ArtificialOwl:
I have no problem merging this, but might be nice to display a message somewhere that external members have no use if public share is disable

ArtificialOwl (approved):

ArtificialOwl:
Circle shares are normal internal shares until you add a mail address as member to the Team. Then a public link is generated and sent by mail

we could disable adding external members (as mail address) to a Team, or displaying a notice that those people wont receive a mail because of the current setup

artonge:
@marcoambrosini, @nfebe, what do you think about Maxence suggestion? It would indeed make sense to add a warning in the sharing sidebar when public shares are disabled.

PR #1875, opened 2025-02-18
feat(psalm): Add all stubs

provokateurin:
I tried to fix a bunch of the errors that were uncovered by adding the proper stubs, but didn't manage to fix all of them since I don't understand everything yet.
I also don't understand why psalm doesn't properly understand that CoreQueryBuilder implements IQueryBuilder, but the change makes it work :woman_shrugging:


ArtificialOwl (approved):

lib/Command/CirclesMemberships.php
```diff
@@ -204,7 +204,7 @@ protected function execute(InputInterface $input, OutputInterface $output): int
 		if ($count === 0) {
 			$output->writeln('(database not updated)');
 		} else {
-			$output->writeln('(' . $count . ' entries generated/updated in the database)');
+			$output->writeln('(' . ((string)$count) . ' entries generated/updated in the database)');
```
come-nc, line 207: Why are all these casts needed, it’s perfectly fine to concatenate an int with a string, no?

provokateurin:
> Why does it need such a recent PHP?

Yeah it's a bit weird that the version requirements are so strict.

come-nc (approved):
I would prefer if the PHP version thing is solved and I can run the analysis locally.
Also the CI will have trouble I expect as it does not always have the last PHP?
And would be better if the IQueryBuilder thing is properly fixed as well.

Not gonna block this PR though, it’s still a step forward.

provokateurin:
I'll try to improve this PR, but I already gave up on the IQueryBuilder thing because there is no trace of what is wrong :woman_shrugging:

PR #1894, opened 2025-03-03
fix(SyncService): Sync group display names

provokateurin:
Follow up to https://github.com/nextcloud/circles/pull/1829.


ArtificialOwl (approved):

provokateurin:
@ArtificialOwl since you started backport https://github.com/nextcloud/circles/pull/1829 down to stable28, this should probably also be backported to all the same versions?

PR #1904, opened 2025-03-11
ignore non-existant users

ArtificialOwl:
verify that user still exists

### DIFF
--- a/lib/Command/MigrateCustomGroups.php
+++ b/lib/Command/MigrateCustomGroups.php
@@ -24,7 +24,7 @@ use Symfony\Component\Console\Output\OutputInterface;
 
 class MigrateCustomGroups extends Base {
 	private OutputInterface $output;
-	/** @var IFederatedUser[] */
+	/** @var array<string, null|IFederatedUser> */
 	private array $fedList = [];
 
 	public function __construct(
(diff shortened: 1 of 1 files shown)


lib/Command/MigrateCustomGroups.php
```diff
@@ -167,11 +179,16 @@ public function updateShares(string $groupUri, string $circleId, array $memberId
...
-	 * @return FederatedUser
+	 * @return null|FederatedUser
 	 */
-	private function cachedFed(string $userId): FederatedUser {
+	private function cachedFed(string $userId): ?FederatedUser {
 		if (!array_key_exists($userId, $this->fedList)) {
-			$this->fedList[$userId] = $this->circlesManager->getLocalFederatedUser($userId);
+			try {
+				$this->fedList[$userId] = $this->circlesManager->getLocalFederatedUser($userId);
+			} catch (\Exception) {
```
provokateurin, line 188: Maybe a bit broad to catch all exceptions here?
At least the exception should be logged, if you make it this broad.

ArtificialOwl: adding the exception to the warning

provokateurin (approved):

PR #1913, opened 2025-03-16
Speedup circles fetching

Koc:
In our organization we have circles with thousands of people. This leads api for fetching collectives list extremely slow, up to 30s. This happens because we loading all members of the circle for `canLeave` check. With suggested change we can reduce amount of loaded users


juliusknorr (approved):

ArtificialOwl:
please set `$limit` to `0` instead of `null` by default and use its value only if `> 0`

There should be better way to implement this, but I am still fine

Koc:
@ArtificialOwl I did requested changes.

Can we merge it?
Should I backport it to v30?

mejo- (approved):

PR #1915, opened 2025-03-18
composer upgrade and cs:fix

ArtificialOwl:


provokateurin:
Is it necessary to upgrade? Not that I'm against it, just wondering...

ArtificialOwl:
not at all, I will revert the composer upgrade but keep the rest of the edit 

_if it's not broken, don't touch it_

Issue #1796, opened 2024-12-05
Duplicate LEFT JOIN when listing members


ChristophWurst:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

### Steps to reproduce
1. ???
2. Look at db logs

### Expected behaviour

Efficient queries.

### Actual behaviour

```sql
...
FROM oc_circles_member c
LEFT JOIN oc_circles_circle c_g ON c_g.unique_id = c.single_id
LEFT JOIN oc_circles_circle c_b ON c_b.unique_id = c.circle_id
LEFT JOIN oc_circles_circle c_o ON c.invited_by = c_o.unique_id
LEFT JOIN oc_circles_member c_b_d ON (c_b_d.circle_id = c_b.unique_id)
AND (c_b_d.level = 9)
LEFT JOIN oc_circles_member c_o_d ON (c_o_d.circle_id = c_o.unique_id)
AND (c_o_d.level = 9)
WHERE (c.circle_id = 'abcdefg')
AND (c.instance = '')
AND (c.level >= 1)
ORDER BY c.level DESC,
c.cached_name ASC
```

^ oc_circles_member is joined **twice**.

### Server configuration

**Web server:** Apache/Nginx

**Database:** MySQL/Maria/SQLite/PostgreSQL

**PHP version:** 8.1/8.2/8.3

**Nextcloud version:** (see Nextcloud admin page)

<details>
<summary>List of activated apps</summary>

```
If you have access to your command line run e.g.:
sudo -u www-data php occ app:list
from within your Nextcloud installation folder
```
</details>

<details>
<summary>Nextcloud configuration</summary>

```
If you have access to your command line run e.g.:
sudo -u www-data php occ config:list system
from within your Nextcloud installation folder
```
</details>

### Browser

**Browser name:** Firefox/Chrome/Safari/…

**Browser version:** 124/125/…

**Operating system:** Windows/Ubuntu/Mac/…

<details>
<summary>Browser log</summary>

```
Insert your browser log here, this could for example include:
a) The javascript console log
b) The network log
c) ...
```

</details>

provokateurin:
Apparently the double left join is necessary, because they don't fetch the same data.

PR #1956, opened 2025-04-07
perf(MemberRequest): Remove duplicate left join

provokateurin:
Closes https://github.com/nextcloud/circles/issues/1796


provokateurin:
Apparently the double left join is necessary, because they don't fetch the same data.

PR #1989, opened 2025-05-12
update child share permissions

ArtificialOwl:


julien-nc:
Tested on a dev setup, works fine. :+1: 
My issue was: the permission change (made by the share owner) were not reflected on the recipient side after the recipient had renamed the shared folder.

julien-nc:
Oops, correction, i can't really test that on 32 because i can't rename the shared directory.

I get
``` xml
<d:error xmlns:d="DAV:" xmlns:s="http://sabredav.org/ns">
  <s:exception>OCP\Files\NotFoundException</s:exception>
  <s:message>Node for share not found, fileid: 728</s:message>
</d:error>
```
for the MOVE request to `remote.php/dav/files/admin/ORIGINAL_PATH` when renaming (from the recipient account).

julien-nc (approved):

Issue #1447, opened 2023-11-30
Mail is send as somebody when using Circles


Gerben-W:
When sharing files with a single mail address, the subject of the mail contains the user that shares this file. 
( USERNAME shared »FILENAME« with you )

But when sharing to a Circle, the subject of the mail contains "someone".
( someone shared a file with you. )

Is it possible to change the mail subject so that the username is in the subject just like sharing a file with a single mail address?

Noki77:
I diagnosed this a fair bit now, since i saw this as well after a sharing attempt of one of our customers bounced...
seems like the issue is caused by `ShareWrapper::hasInitiator` returning `false` in [ShareCreatedSendMail.php#L165](https://github.com/nextcloud/circles/blob/master/lib/Listeners/Files/ShareCreatedSendMail.php#L165), though I haven't figured out why yet...
In my case I "fixed" the issue by simply replacing "someone" within the mail subject by using an event listener on the  `BeforeMessageSent` Event in a custom NextCloud App, that i'm using anyway for having certain email templates based on the user mail address or groups.

So if you still want to "fix" this yourself and you're willing to go through the effort of making an own NC app for it, here is basically what you need to do:
* Create a custom app (just follow the [docs](https://docs.nextcloud.com/server/latest/developer_manual//getting_started/index.html) or use the [app skeleton generator by NextCloud](https://apps.nextcloud.com/developer/apps/generate))
* write an event listener on `\OCA\Circles\Events\AddingCircleMemberEvent`, that fetches the circle owner and stores it somewhere (in my case I have an `Application::setDispatchingUser` function
* write an event listener on `OCP\Mail\Events\BeforeMessageSent` that first checks whether a circle owner was found earlier (in my case just check whether `Application::getDispatchingUser()` is not null), then whether `$event->getMessage()->getSubject()` contains  the word "someone" (yes, IMessage doesn't have a `getSubject()` function, but `\OC\Mail\Message` does, maybe check if it's an instance of it first) and lastly replace "someone" with the previously fetched circle owners' display name
* register both listeners in your `Application::register` method

Gerben-W:
@Noki77 Thanks for helping with this.
Loved your solution and thinking of implementing this.
Unfortunately I only have one concern you might be able to help with.
In the mail which is send out, the body contains the Circle name, which can be anything (even non positive names to group a certain amount of people). Do you know how to easily replace the circle name with the word "you" in the mail that is send out.
[...]

PR #2007, opened 2025-05-28
fix: use shared_by user in share mail if initiator is not set

icewind1991:
Fixes https://github.com/nextcloud/circles/issues/1447


ArtificialOwl (approved):

provokateurin (changes requested):
Please revert the cs-fixer changes and rebase on master. The issue with it in the CI has been solved: https://github.com/nextcloud/circles/pull/2043
Other than that this looks good, just needs a squash.

PR #2012, opened 2025-06-02
fix(ShareByCircleProvider): Include data in getSharesByPath to fix share_with_displayname

provokateurin:
Fixes https://github.com/nextcloud/server/issues/51726


provokateurin:
stable30 should not suffer from this bug as the code path is not triggered there.

julien-nc (approved):
Tested on server's master and stable31. Works just fine! :+1:

PR #2050, opened 2025-06-30
feature: send email to internal users of circles when shared with circle

yemkareems:
feature: send email to internal users of circles when shared with circle


lib/Service/CircleShareMailerService.php
```diff
@@ -0,0 +1,143 @@
+<?php
+
+declare(strict_types=1);
+/**
+ * SPDX-FileCopyrightText: 2016 Nextcloud GmbH and Nextcloud contributors
```
provokateurin: ?

lib/Service/CircleShareMailerService.php
```diff
@@ -0,0 +1,143 @@
...
+		Defaults $defaults,
+	) {
+		$this->mailer = $mailer;
+		$this->l = $l10nFactory->get('circles');
+		$this->logger = $logger;
+		$this->urlGenerator = $urlGenerator;
+		$this->config = $config;
+		$this->userManager = $userManager;
+		$this->defaults = $defaults;
+	}
```
provokateurin: Use constructor property promotion.

provokateurin (commented):
Isn't this kind of logic supposed to be implemented through IShareProviderWithNotification?

lib/Service/CircleShareMailerService.php
yemkareems: Sure will refactor this to a Provider

provokateurin: This comment specifically was about the license year

yemkareems: Sorry about the confusion. Will make it 2025

lib/ShareByCircleProvider.php
```diff
@@ -143,6 +145,11 @@ public function create(IShare $share): IShare {
 		$share->setToken($this->token(15));
 		$owner = $circle->getInitiator();
 		$this->shareWrapperService->save($share);
+		try {
+			$this->circleShareMailProvider->sendShareNotification($share, $circle);
+		} catch (Exception $e) {
+			$this->logger->error('sending sharing email failed', [$e->getMessage()]);
+		}
```
provokateurin: This should be handled by https://github.com/nextcloud/server/blob/d55b1eeb33c9752c7e9075f26c2a0c7493680080/lib/private/Share20/Manager.php#L750-L764 automatically.

lib/Provider/CircleShareMailProvider.php
```diff
@@ -0,0 +1,131 @@
...
+use OCP\IL10N;
+use OCP\IURLGenerator;
+use OCP\IUser;
+use OCP\IUserManager;
+use OCP\Mail\IMailer;
+use OCP\Share\IShare;
+use OCP\Share\IShareProviderWithNotification;
+use Psr\Log\LoggerInterface;
+
+class CircleShareMailProvider extends DefaultShareProvider implements IShareProviderWithNotification {
```
provokateurin: You are not implementing sendMailNotification now

provokateurin (changes requested):

lib/ShareByCircleProvider.php
yemkareems: Changed ShareByCircleProvider to implement IShareProviderWithNotification to send automatically. For this share mail_send had to be set to true to work properly which i did while share creation and in wrapper also.

lib/Provider/CircleShareMailProvider.php
yemkareems: sendMailNotification is implemented in ShareByCircleProvider and this is renamed to a helper to send email. link in email is public share link with share token

lib/Helpers/CircleShareMailHelper.php
provokateurin, line 1: This logic should just be moved into the ShareByCircleProvider

yemkareems: Moved helper functions to provider

lib/ShareByCircleProvider.php
```diff
@@ -161,6 +170,127 @@ public function create(IShare $share): IShare {
...
+		} catch (Exception $e) {
+			//fail silently as share created already and mail sending alone failed
+			$this->logger->error('Circle share internal email sending failed', [$e->getMessage()]);
+		}
+
+		return false;
+	}
+
+	public function sendShareNotification(IShare $share, $circle): void {
+		if ($this->config->getSystemValueBool('sharing.enable_share_mail', true)) {
```
provokateurin:
```suggestion
```
This check is already done by the share manager before sendMailNotification is called.

provokateurin (changes requested):
Woops I forgot to submit my review

lib/ShareByCircleProvider.php
yemkareems: Removed this check

lib/ShareByCircleProvider.php
```diff
@@ -161,6 +169,125 @@ public function create(IShare $share): IShare {
 		return $wrappedShare->getShare($this->rootFolder, $this->userManager, $this->urlGenerator);
 	}
 
+	/**
+	 * @inheritDoc
+	 */
+	public function sendMailNotification(IShare $share): bool {
```
provokateurin:
```suggestion
	public function sendMailNotification(IShare $share): bool {
```

Not necessary, this is already the default if there is no comment.

lib/ShareByCircleProvider.php
```diff
@@ -161,6 +169,125 @@ public function create(IShare $share): IShare {
...
+		$dataProbe->add(DataProbe::OWNER)
+			->add(DataProbe::INITIATOR, [DataProbe::BASED_ON]);
+
+		$circle = $this->circleService->probeCircle($share->getSharedWith(), $circleProbe, $dataProbe);
+		try {
+			$this->sendShareNotification($share, $circle);
+			return true;
+		} catch (Exception $e) {
+			//fail silently as share created already and mail sending alone failed
+			$this->logger->error('Circle share internal email sending failed', [$e->getMessage()]);
```
provokateurin:
```suggestion
			$this->logger->error('Failed to send email for circle share', [$e->getMessage()]);
```

lib/ShareByCircleProvider.php
```diff
@@ -161,6 +169,125 @@ public function create(IShare $share): IShare {
...
+		return false;
+	}
+
+	public function sendShareNotification(IShare $share, $circle): void {
+		$circleMembers = $circle->getMembers();
+		$link = $this->urlGenerator->linkToRouteAbsolute('files_sharing.sharecontroller.showShare', [
+			'token' => $share->getToken()
+		]);
+		foreach ($circleMembers as $member) {
+			if ($member->getUserType() != 1) {
```
provokateurin:
```suggestion
			if ($member->getUserType() != Member::TYPE_USER) {
```

lib/ShareByCircleProvider.php
```diff
@@ -161,6 +169,125 @@ public function create(IShare $share): IShare {
...
+				continue;
+			}
+			$user = $this->userManager->get($member->getUserId());
+			if ($user !== null) {
+				$email = $user->getEMailAddress();
+				if ($email != '' && $this->mailer->validateMailAddress($email)) {
```
provokateurin: Please use early-continue to avoid such deep nesting.

lib/ShareByCircleProvider.php
```diff
@@ -161,6 +169,125 @@ public function create(IShare $share): IShare {
...
+					$this->sendUserShareMail(
+						$this->l10n,
+						$share->getNode()->getName(),
+						$link,
+						$share->getSharedBy(),
+						$email,
+						$share->getExpirationDate(),
+						$share->getNote()
```
provokateurin:
```suggestion
						$share->getNode()->getName(),
						$link,
						$share->getSharedBy(),
						$email,
						$share->getExpirationDate(),
						$share->getNote()
```
This doesn't need to be passed as the other method can access it as well.

provokateurin:
```suggestion
						$this->l10n,
						$share,
						$link,
						$email,
```
Just passing the entire share is a lot easier.

lib/ShareByCircleProvider.php
```diff
@@ -161,6 +169,125 @@ public function create(IShare $share): IShare {
...
+	}
+
+	protected function sendUserShareMail(
+		IL10N $l,
+		$filename,
+		$link,
+		$initiator,
+		$shareWith,
+		?\DateTime $expiration = null,
+		$note = '',
```
provokateurin: This should have strong types.

lib/ShareByCircleProvider.php
```diff
@@ -161,6 +169,125 @@ public function create(IShare $share): IShare {
...
+			return true;
+		} catch (Exception $e) {
+			//fail silently as share created already and mail sending alone failed
+			$this->logger->error('Circle share internal email sending failed', [$e->getMessage()]);
+		}
+
+		return false;
+	}
+
+	public function sendShareNotification(IShare $share, $circle): void {
```
provokateurin:
```suggestion
	public function sendShareNotification(IShare $share, Circle $circle): void {
```

lib/ShareByCircleProvider.php
```diff
@@ -161,6 +169,125 @@ public function create(IShare $share): IShare {
...
+		IL10N $l,
+		$filename,
+		$link,
+		$initiator,
+		$shareWith,
+		?\DateTime $expiration = null,
+		$note = '',
+	): void {
+		$initiatorUser = $this->userManager->get($initiator);
+		$initiatorDisplayName = ($initiatorUser instanceof IUser) ? $initiatorUser->getDisplayName() : $initiator;
```
provokateurin: This feels quite dangerous, there could easily be an overlap of the getSharedBy value matching a uid by accident.

lib/ShareByCircleProvider.php
```diff
@@ -161,6 +169,125 @@ public function create(IShare $share): IShare {
...
+						$email,
+						$share->getExpirationDate(),
+						$share->getNote()
+					);
+				}
+			}
+		}
+	}
+
+	protected function sendUserShareMail(
```
provokateurin: I guess this was all copied from DefaultShareProvider?

provokateurin (changes requested):

lib/ShareByCircleProvider.php
yemkareems: Removed comment

lib/ShareByCircleProvider.php
yemkareems: Corrected text as mentioned

lib/ShareByCircleProvider.php
yemkareems: if check clubbed with continue

lib/ShareByCircleProvider.php
yemkareems: Removed l10n

yemkareems: Passed share and removed other related params

lib/ShareByCircleProvider.php
yemkareems: Type hinting for circle done

lib/ShareByCircleProvider.php
yemkareems: Removed few params and type hinting of other params done

lib/ShareByCircleProvider.php
yemkareems: Constant used :)

lib/ShareByCircleProvider.php
yemkareems: Yes it was done to not break other functionality

lib/ShareByCircleProvider.php
```diff
@@ -161,6 +171,125 @@ public function create(IShare $share): IShare {
...
+			'token' => $share->getToken()
+		]);
+		foreach ($circleMembers as $member) {
+			$user = $this->userManager->get($member->getUserId());
+
+			if ($member->getUserType() !== Member::TYPE_USER
+				|| $user === null
+				|| !$this->mailer->validateMailAddress($user->getEMailAddress())) {
+				continue;
+			}
```
provokateurin:
```suggestion
			if ($member->getUserType() !== Member::TYPE_USER) {
				continue;
			}
			
			$user = $this->userManager->get($member->getUserId());
			if ($user === null) {
				continue;
			}

			if (!$this->mailer->validateMailAddress($user->getEMailAddress())) {
				continue;
			}
```

To avoid querying the user manager if the member isn't even a user

lib/ShareByCircleProvider.php
```diff
@@ -161,6 +171,125 @@ public function create(IShare $share): IShare {
...
+		$circleMembers = $circle->getMembers();
+		$link = $this->urlGenerator->linkToRouteAbsolute('files_sharing.sharecontroller.showShare', [
+			'token' => $share->getToken()
+		]);
+		foreach ($circleMembers as $member) {
+			$user = $this->userManager->get($member->getUserId());
+
+			if ($member->getUserType() !== Member::TYPE_USER
+				|| $user === null
+				|| !$this->mailer->validateMailAddress($user->getEMailAddress())) {
```
provokateurin: Is it actually necessary to validate the email here? I'd think that it was already validated when it is set by the user, but maybe you can explain this to me?

lib/ShareByCircleProvider.php
```diff
@@ -161,6 +171,125 @@ public function create(IShare $share): IShare {
...
+	 *
+	 * @param string $link link to the file/folder
+	 * @param string $shareWith email address of share receiver
+	 * @param IShare $share
+	 * @throws \Exception
+	 */
+	protected function sendUserShareMail(
+		string $link,
+		string $shareWith,
+		IShare $share): void {
```
provokateurin:
```suggestion
		IShare $share,
	): void {
```
Better readability and avoids annoying diff if another parameter is added later.

lib/ShareByCircleProvider.php
provokateurin: I guess it is fine, I don't see a way to fix it somehow :woman_shrugging:

lib/ShareByCircleProvider.php
```diff
@@ -161,6 +171,125 @@ public function create(IShare $share): IShare {
...
+
+		$filename = $share->getNode()->getName();
+		$initiator = $share->getSharedBy();
+		$expiration = $share->getExpirationDate();
+		$note = $share->getNote();
+		$l = $this->l10n;
+
+		$initiatorUser = $this->userManager->get($initiator);
+		$initiatorDisplayName = ($initiatorUser instanceof IUser) ? $initiatorUser->getDisplayName() : $initiator;
```
provokateurin: Please do these outside of the method, since it will be called in a loop and just pass the $initiatorUser to the method.

lib/Db/ShareWrapperRequest.php
```diff
@@ -42,6 +42,8 @@ public function save(IShare $share, int $parentId = 0): int {
 		$qb = $this->getShareInsertSql();
 		$qb->setValue('attributes', $qb->createNamedParameter($this->formatShareAttributes($share->getAttributes())))
 			->setValue('share_type', $qb->createNamedParameter($share->getShareType()))
+			->setValue('mail_send', $qb->createNamedParameter($share->getMailSend()))
+			->setValue('note', $qb->createNamedParameter($share->getNote()))
```
provokateurin, line 46: Is this actually necessary? While the note is needed in the email that is sent, this is about setting the note from the share when it is created (which is either unnecessary or a separate fix).

provokateurin (dismissed):

lib/ShareByCircleProvider.php
yemkareems: There maybe cases when the user is created without a email address

lib/Db/ShareWrapperRequest.php
yemkareems: If the note is not set here the email is missing the note and hence this is required

lib/Listeners/Files/ShareCreatedSendMail.php
```diff
@@ -148,5 +159,42 @@ public function handle(Event $event): void {
 				);
 			}
 		}
+
+		$circleMembers = $circle->getMembers();
```
ArtificialOwl: You should use `getInheritedMembers()` so that you support inheritance in case other circle(s) is member of that circle

In fact, it might be interesting to use the previous `foreach ($circle->getInheritedMembers(false, true) as $member)` and add your code, or a method to your code, there (before the _if_ condition):
https://github.com/nextcloud/circles/blob/a9a459ca23301e55efd3bfd5dfb96c166a5db689/lib/Listeners/Files/ShareCreatedSendMail.php#L110-L112

lib/Listeners/Files/ShareCreatedSendMail.php
```diff
@@ -148,5 +159,42 @@ public function handle(Event $event): void {
...
+				continue;
+			}
+			$email = $user->getEMailAddress();
+			if ($email === null
+				|| $email === $initiatorEmail
+			) {
+				continue;
+			}
+
+			$this->sendMailService->sendUserShareMail(
```
ArtificialOwl: have you checked if `generateMail()` cannot do the same _or almost_ ?

https://github.com/nextcloud/circles/blob/bb52de40f3f46dd97fbd5289d1180dbc17409507/lib/Service/SendMailService.php#L82-L89

lib/Listeners/Files/ShareCreatedSendMail.php
```diff
@@ -148,5 +159,42 @@ public function handle(Event $event): void {
...
+		$link = $this->urlGenerator->linkToRouteAbsolute('files_sharing.sharecontroller.showShare', [
+			'token' => $share->getToken()
+		]);
+		$initiator = $share->getSharedBy();
+		$initiatorUser = $this->userManager->get($initiator);
+		$initiatorDisplayName = ($initiatorUser instanceof IUser) ? $initiatorUser->getDisplayName() : $initiator;
+		$initiatorEmail = ($initiatorUser instanceof IUser) ? $initiatorUser->getEMailAddress() : null;
+
+		foreach ($circleMembers as $member) {
+			if ($member->getUserType() !== Member::TYPE_USER) {
```
ArtificialOwl: also check that `$member->isLocal()`

lib/Listeners/Files/ShareCreatedSendMail.php
yemkareems: Added user check above this if condition

lib/Listeners/Files/ShareCreatedSendMail.php
yemkareems: sendUserShareMail and generateMail are using different share objects and so keeping it seperate

lib/Listeners/Files/ShareCreatedSendMail.php
yemkareems: Added the isLocal check to type user check

Issue #1938, opened 2025-03-29
👥 Teams - Roadmap for NC32


AndyScherzinger:
We want to improve the usability of the Teams UI. There are several enhancements planned for the Teams UI that have been broken down into the subtasks below.

## Dashboard Widget changes

<img width="495" height="754" alt="Image" src="https://github.com/user-attachments/assets/3090ed9e-df64-412c-8af0-a2469c570e27" />

## Teams management UI changes

| Non-Member | Member | Admin |
| ---------- | ------ | ----- |
|  ![Image](https://github.com/user-attachments/assets/0de14241-c1d8-430e-b7d6-8f7f22031a71) | ![Image](https://github.com/user-attachments/assets/da0fbe8b-06d6-42ad-b212-1a7e5dda8d87) | ![Image](https://github.com/user-attachments/assets/f3425920-28b4-4ee5-9bf9-b1078ea6d1e8) |

## Management

| Editing | Settings |
| ------- | -------- |
| ![Image](https://github.com/user-attachments/assets/73dfa848-6543-452a-ac51-cd0898b64795) | ![Image](https://github.com/user-attachments/assets/be748373-e6ef-4715-a7ae-aee6cc1955fc) |

ArtificialOwl:
Quick summary on those tasks

- [ACL editing for teams #1939](https://github.com/nextcloud/circles/issues/1939)
  Back-end part should be over, needs front-end stuff

- [Ability to restrict teams creation to defined group(s) / admins #1940](https://github.com/nextcloud/circles/issues/1940)
  There is a config value in the circles app that allows to limit the creation of Teams to members of a specific Team. Because users, groups and teams can be added to that team, it fulfill this request. Just need it to be documented

- [Switch to default-expose groups as "read-only" teams #1941](https://github.com/nextcloud/circles/issues/1941)
  I need details on this one
  
- [Double-size Dashboard Widget - all teams + assigned resources (design spec) + "plus"-button leading to Teams interface #1942](https://github.com/nextcloud/circles/issues/1942)
   Sound like a front-end thing ?

- [Improved frontend for teams management in Contacts UI #1944](https://github.com/nextcloud/circles/issues/1944)
   This sounds like a design/front-end assignation

-  [Federated teams (eventually, rather 2026) #1943](https://github.com/nextcloud/circles/issues/1943)
    This is the big thing I will work on for 32. The feature exists in a PoC state and needs some improvement :
       * easier setup and test
       * documentation
[...]

Issue #1942, opened 2025-03-29
New dashboard teams widget


AndyScherzinger:
Planned changes (to be edited if necessary)

## New dashboard widget

<img width="495" height="754" alt="Image" src="https://github.com/user-attachments/assets/e6555cc2-38dc-4007-a020-6bb7d7f08062" />

marcoambrosini:
@jancborchardt and I discussed this a bit and came to the conclusion that something like replicating the team UI here would be ideal.

The problem I see, on second thought, is that we cannot have a scroll container within this the dashboard, because the dashboard itself is a scroll container

If the goal here is team management without diving into contacts, I think we're better off bringing it up a dialog. With a dialog we can block the scroll container behind, so that's not a limitation anymore.

@AndyScherzinger do you remember the root issue that originated this idea of the double width widget?

jancborchardt:
@marcoambrosini just like with other widgets, we can show the top 6/7 teams, and their contents on the right.

For anything else on the bottom would be a "Show more Teams" button, and "Show team details" on the right side.

marcoambrosini:
I don't think this would look good or be particularily useful though, but if that's exactly what's needed/wanted I can write down a spec or a quick mockup

marcoambrosini:
If the blur filter issue is solved we could also see how it looks if we blur the background

jancborchardt:
Looks really really nice, as discussed @kra-mo @marcoambrosini! :) Good on fixing some of the details directly. I’m excited to see the interaction, but as said if it doesn’t work out, I think a more barebones version without the focus and scrolling (but instead the "Show more" buttons) is ok as well. Of course not as polished, but not breaking.

And as discussed it would be cool to just have a mockup of a 2-width widget just for reference / comparison. :)

PR #2085, opened 2025-08-16
feat: Add interactive vue-based teams dashboard widget

nfebe:
| Before | After |
| ---------- | ------- |
| image | image |

### Video

https://github.com/user-attachments/assets/f140cefa-f652-462d-8c08-15b5236fec44

Resolves: https://github.com/nextcloud/circles/issues/1938


nfebe (commented):
## Notes
- [ ] Since we now used vue-based widget, it could be possible to move the js code to the contacts app
- [ ] Based on questions in : https://github.com/nextcloud/circles/issues/1942#issuecomment-3183448039 we could add tooltips that provide more context on file hovers.

AndyScherzinger:
Looks good to me, just some minor comments to be checked with @jancborchardt

In the last call we decided to have the app icon on the bottom right of each resource to know its origin

The original mockup hat a right scroll ability on the list the above video shows a '+X' items visual instead which should be approved by designers.

In any way of doable as a quick follow up PR I'd be fine to merge.

AndyScherzinger (approved):

susnux (changes requested):
This needs a build system for the vue components - the circles app is not available in Nextcloud build time!
(Also it should be then moved to the `src` directory rather then `js`)

susnux:
*(Took over due to vacations)*

Added the build system:
- Add basic vite based build system
- Added ESLint workflow
- Added Typescript setup
- Fixed linter issues
- Added necessary CI workflows

Additionally to that I also used directly Vue 3 instead.

Content wise only following is changed:
- The changes requested by @kra-mo are now implemented (spacing and wording).
- As well as the button is secondary as requested.

lib/Service/TeamResourceService.php
```diff
@@ -0,0 +1,72 @@
+<?php
+
+/**
+ * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
```
artonge:
```suggestion
 * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
```

src/components/TeamResources.vue
```diff
@@ -0,0 +1,129 @@
...
+	&__list {
+		display: flex;
+		align-items: center;
+		flex-wrap: wrap;
+		gap: var(--resources-list-gap);
+		list-style: none;
+	}
+
+	&__box {
+		cursor: pointer;
```
artonge: Is this needed? I don't see any `click` event listener on `box`. The `<a>` should be enough, no?

artonge (approved):

src/components/TeamResources.vue
susnux: Was already done in the original code - did not change those rules - but I agree!

susnux:
```suggestion
```

PR #2090, opened 2025-08-19
fix(share): returns expiration date to front-end

ArtificialOwl:
fix the display of expiration date on existing share

### DIFF
--- a/lib/Model/ShareWrapper.php
+++ b/lib/Model/ShareWrapper.php
@@ -32,6 +32,7 @@ use OCP\Server;
 use OCP\Share\Exceptions\IllegalIDChangeException;
 use OCP\Share\IAttributes;
 use OCP\Share\IShare;
+use Psr\Log\LoggerInterface;
 
 /**
  * Class ShareWrapper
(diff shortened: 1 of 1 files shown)


yemkareems (approved):

lib/Model/ShareWrapper.php
```diff
@@ -570,6 +573,11 @@ public function importFromDatabase(array $data, string $prefix = ''): IQueryRow
 			->setShareTime($shareTime)
 			->setShareNote($this->get($prefix . 'note', $data));
 
+		try {
+			$this->setExpirationDate(new DateTime($this->get('expiration', $data)));
+		} catch (\Exception $e) {
+		}
```
provokateurin: Silent catch is bad

ArtificialOwl: expiration can be null

ArtificialOwl: i'll add a debug

provokateurin: Then detect that case or catch the specific exception please

PR #2108, opened 2025-09-04
feat: caching user single id in preferences

ArtificialOwl:
- local account single id is now stored in preferences,
- local cache instead of distributed for other entities,

- needs to be backported to 30,
- backport will not use new UserConfig+Lexicon


lib/Service/FederatedUserService.php
```diff
@@ -192,7 +195,7 @@ public function __construct(
 		$this->interfaceService = $interfaceService;
 		$this->configService = $configService;
 
-		$this->cache = $cacheFactory->createDistributed(self::CACHE_SINGLE_CIRCLE);
+		$this->cache = $cacheFactory->createLocal(self::CACHE_SINGLE_CIRCLE);
```
provokateurin: Why?

lib/Service/FederatedUserService.php
```diff
@@ -520,11 +523,44 @@ public function getLocalFederatedUser(string $userId, bool $check = true, bool $
...
+	/**
+	 * get singleId from UserConfig (already loaded from database)
+	 * and emulate FederatedUser for local accounts.
+	 */
+	private function getCachedLocalFederatedUser(IFederatedUser $federatedUser): bool {
+		if ($federatedUser->getUserType() !== Member::TYPE_USER || !$federatedUser->isLocal()) {
+			return false;
+		}
+
+		$teamSingleId = $this->userConfig->getValueString($federatedUser->getUserId(), Application::APP_ID, ConfigLexicon::USER_SINGLE_ID);
```
provokateurin:
```suggestion
		$userSingleId = $this->userConfig->getValueString($federatedUser->getUserId(), Application::APP_ID, ConfigLexicon::USER_SINGLE_ID);
```
?

lib/Service/FederatedUserService.php
ArtificialOwl: cached data is static and have no interest being shared, therefor local should be better than distributed

I might be in the wrong here, if local is not lighter than distributed I can revert this change

provokateurin: Well it can be faster, but with a distributed cache the cache hit rate might be higher.

ArtificialOwl: ok, i'll revert to distributed

provokateurin (approved):

PR #2112, opened 2025-09-07
feat: add search and limit support for team members list

AIlkiv:
This will enable us to optimize the retrieval of participant lists for large teams in the Contacts and Collectives apps.

This PR extends the API. Therefore, please backport it to 32, 31 so that it can be used for other modules that still support version 32, 31.


mejo- (approved):
Changes look sensible to me.

PR #2157, opened 2025-10-08
chore: Fix typings from parent class

juliusknorr:
Fixes current master so you can enable circles again without having a broken instance

### DIFF
--- a/lib/MountManager/CircleMountManager.php
+++ b/lib/MountManager/CircleMountManager.php
@@ -71,21 +71,6 @@ class CircleMountManager extends Manager {
 		return true;
 	}
 
-
-	// TODO: implement !
-	public function getMountManager() {
-		return $this;
-	}
-
-	// TODO: implement !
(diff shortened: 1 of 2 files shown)


CarlSchwan (approved):

lib/MountManager/CircleMountManager.php
```diff
@@ -73,7 +73,7 @@ public function renameShare(int $gsShareId, string $target) {
 
 
 	// TODO: implement !
-	public function getMountManager() {
+	public function getMountManager(): \OC\Files\Mount\Manager {
 		return $this;
```
juliusknorr: @CarlSchwan @ArtificialOwl I'm not sure the implementation here is actually correct or if it should rather return the files mount manager from parent?

CarlSchwan: actually maybe this should just be removed and inherit from the external share manager.

PR #2167, opened 2025-10-14
refactor(QueryBuilder): Migrate away from deprecated execute method

mejo-:
Required due to https://github.com/nextcloud/server/pull/55720

### DIFF
--- a/lib/Db/TokensRequest.php
+++ b/lib/Db/TokensRequest.php
@@ -8,7 +8,7 @@
 
 namespace OCA\Circles\Db;
 
-use Doctrine\DBAL\Exception\UniqueConstraintViolationException;
+use OC\DB\Exceptions\DbalException;
 use OCA\Circles\Exceptions\TokenDoesNotExistException;
 use OCA\Circles\Model\DeprecatedMember;
 use OCA\Circles\Model\SharesToken;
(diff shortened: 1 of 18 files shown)


lib/Db/TokensRequest.php
```diff
@@ -114,7 +114,7 @@ public function generateTokenForMember(DeprecatedMember $member, int $shareId, s
 				->setValue('token', $qb->createNamedParameter($token))
 				->setValue('password', $qb->createNamedParameter($password));
 
-			$qb->execute();
+			$qb->executeStatement();
 		} catch (UniqueConstraintViolationException $e) {
```
ChristophWurst: As far as I remember the exception handling is not identical for `execute` and `executeStatement`. Please double-check :)

mejo-: Indeed `ConnectionAdapter::executeStatement()` throws original exceptions wrapped into `DbalException`, see https://github.com/nextcloud/server/blob/1eff9e5c905d2472179dddcd87a5b911e0686970/lib/private/DB/ConnectionAdapter.php#L68.

Good catch @ChristophWurst! Adjusted now.

lib/Db/TokensRequest.php
```diff
@@ -8,7 +8,7 @@
 
 namespace OCA\Circles\Db;
 
-use Doctrine\DBAL\Exception\UniqueConstraintViolationException;
+use OC\DB\Exceptions\DbalException;
```
nickvergessen, line 11: Should have been the public class:
```suggestion
use OCP\DB\Exceptions;
```
https://github.com/nextcloud/server/blob/master/lib/public/DB/Exception.php#L121

mejo-: Thanks, fixed here: https://github.com/nextcloud/circles/pull/2171

PR #2172, opened 2025-10-15
fix: Fix left-over execute

CarlSchwan:


mejo- (approved):
Oups, good catch. Thank you 🙏

PR #2173, opened 2025-10-15
chore(psalm): Warn about deprecated stuff

CarlSchwan:


tests/psalm-baseline.xml
```diff
@@ -5,7 +5,32 @@
       <code><![CDATA[str_starts_with($sK, '_')]]></code>
     </TypeDoesNotContainType>
   </file>
+  <file src="lib/AppInfo/Application.php">
+    <DeprecatedInterface>
+      <code><![CDATA[$context->getAppContainer()]]></code>
```
nickvergessen, line 10: If you add all this to the baseline, you can also simply not enable the checking.
The next PR in server to remove one of these will still break the app and there was no warning visible as a todo list item?

CarlSchwan: It's mostly to prevent the usage of more deprecated stuff. I already fixed some of the easy fix, but some are more tricky to fix.

lib/Tools/Traits/TNCLocalSignatory.php
```diff
@@ -63,15 +63,15 @@ public function createSimpleSignatory(NCSignatory $signatory): void {
...
+			json_decode(Server::get(IAppConfig::class)->getValueString($app, 'key_pairs', '[]'), true);
 		$signatories[$signatory->getId()] = [
 			'keyId' => $signatory->getKeyId(),
 			'keyOwner' => $signatory->getKeyOwner(),
 			'publicKey' => $signatory->getPublicKey(),
 			'privateKey' => $signatory->getPrivateKey()
 		];
 
-		Server::get(IConfig::class)->setAppValue($app, 'key_pairs', json_encode($signatories));
+		Server::get(IAppConfig::class)->setValueString($app, 'key_pairs', json_encode($signatories));
```
come-nc: getValueArray/setValueArray does the json encoding for you, no?

PR #2174, opened 2025-10-16
refactor: Port away from some deprecated OCP usages

CarlSchwan:


lib/AppInfo/Application.php
```diff
@@ -129,25 +129,19 @@ public function register(IRegistrationContext $context): void {
...
-			->get(ConfigService::class);
-
 		$context->injectFn(Closure::fromCallable([$this, 'registerMountProvider']));
 	}
 
 
-	public function registerMountProvider(IServerContainer $container) {
+	public function registerMountProvider(ContainerInterface $container) {
+		$this->configService = $container->get(ConfigService::class);
 		if (!$this->configService->isGSAvailable()) {
```
come-nc:
```suggestion
		$configService = $container->get(ConfigService::class);
		if (!$configService->isGSAvailable()) {
```
It’s only used here I suppose?
It should be a local var then.

CarlSchwan:
ping

Issue #2181, opened 2025-10-18
Allow Admins to change the name of the Team


janbaum:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

## Feature request

**Which Nextcloud Version are you currently using:** 33.0.0

**Is your feature request related to a problem? Please describe.**
Only owners can rename/change the name of a Team.

**Describe the solution you'd like**
 Admins should be allowed to rename a Team, too

**Describe alternatives you've considered**
> Another option would be to create an own API endpoint in Collectives for renaming a team and a `renameCircle()` function in our `CircleHelper` class by using [`startSuperSession()`](https://github.com/nextcloud/collectives/blob/7afc43903f67803fc00803ae3b801869671f906a/lib/Service/CircleHelper.php#L91). 

 _Originally posted by @mejo- in [#1716](https://github.com/nextcloud/collectives/issues/1716#issuecomment-3396350184)_

**Additional context**
https://github.com/nextcloud/collectives/issues/1716

PR #2190, opened 2025-10-29
allow admins to edit circle details

janbaum:
It should be possible for admins to edit it's name, description and display name. Again this would be necessary for Collectives.

The `verify()` function is only called for editing those three properties.

fixes: #2181


ArtificialOwl:
because this request should only be valid from the owner, we might needs to add some check on federated teams.
i will have a look into this as soon as possible

PR #2197, opened 2025-11-04
refactoring activities

ArtificialOwl:
feed Activities with minimum data

- previous parser is moved in Deprecated folder,
- minimum data is now extracted from circles and members,
- version added to the data,
- new data is using the new parser


artonge (approved):
Why do we need to keep the deprecated ones?

PR #2254, opened 2025-11-29
feat: Add the possibility to create invitation links to circles

Koc:
This PR adds possibility to generate invitation link (even for completely hidden/closed Circles). Any user is able to join to the circle by opening this link.

## :framed_picture: Screenshots

image

image

image

## Dependencies
- ~~has to be merged before https://github.com/nextcloud/contacts/pull/4887~~ not mandatory anymore as we have FE in same module now


lib/Model/CircleInvitation.php
```diff
@@ -0,0 +1,178 @@
...
+	 */
+	public function setCreatedBy(string $createdBy): self {
+		$this->createdBy = $createdBy;
+
+		return $this;
+	}
+
+	/**
+	 * @return string
+	 */
```
artonge: Phpdoc comments are not needed when php types are used, can you remove them? Same thing for all the other additions.

lib/FederatedItems/CircleCreateInvitation.php
```diff
@@ -0,0 +1,90 @@
...
+	 * @throws RequestBuilderException
+	 * @throws CircleNameTooShortException
+	 */
+	public function verify(FederatedEvent $event): void {
+		$circle = $event->getCircle();
+
+		$initiatorHelper = new MemberHelper($circle->getInitiator());
+		$initiatorHelper->mustBeAdmin();
+
+		$new = clone $circle;
```
artonge, line 37: Why is it needed to clone the circle object?

lib/Db/CircleInvitationRequest.php
```diff
@@ -0,0 +1,55 @@
...
+			->setValue('created_by', $qb->createNamedParameter($circleInvitation->getCreatedBy()));
+		$qb->executeStatement();
+	}
+
+	/**
+	 * @param CircleInvitation $circleInvitation
+	 *
+	 * @throws InvalidIdException
+	 */
+	public function replace(CircleInvitation $circleInvitation): void {
```
artonge, line 39: Could there be multiple invitation per circle? If so, shouldn't we delete a specific invitation, and not all of them?

lib/Db/CircleInvitationRequest.php
```diff
@@ -0,0 +1,55 @@
...
+		$this->delete($circleInvitation->getCircleId());
+		$this->save($circleInvitation);
+	}
+
+	/**
+	 * @param string $circleId
+	 */
+	public function delete(string $circleId): void {
+		$qb = $this->getCircleInvitationDeleteSql();
```
artonge: We should inline the content of `getCircleInvitationDeleteSql` and delete it. I know it might not be aligned with the current code base, but we are moving toward reducing the amount of abstraction layer.

lib/Db/CircleRequest.php
```diff
@@ -167,13 +167,17 @@ public function getCircles(?IFederatedUser $initiator, CircleProbe $probe): arra
 			$qb->limitToInitiator(CoreQueryBuilder::CIRCLE, $initiator);
 			$qb->orderBy($qb->generateAlias(CoreQueryBuilder::CIRCLE, CoreQueryBuilder::INITIATOR) . '.level', 'desc');
 			$qb->addOrderBy(CoreQueryBuilder::CIRCLE . '.display_name', 'asc');
+			$qb->leftJoinCircleInvitation(CoreQueryBuilder::CIRCLE);
```
artonge, line 162: Shouldn't we join only if `$probe->hasInvitationCode() === true`?

lib/Db/CircleRequestBuilder.php
```diff
@@ -33,6 +33,16 @@ protected function getCircleInsertSql(): CoreQueryBuilder {
 		return $qb;
 	}
 
+	/**
+	 * @return CoreQueryBuilder&IQueryBuilder
+	 */
+	protected function getCircleInvitationInsertSql(): CoreQueryBuilder {
```
artonge: Similarly to `getCircleInvitationDeleteSql`, we should inline the content of this method.

lib/Service/CircleService.php
```diff
@@ -391,6 +448,11 @@ public function circleJoin(string $circleId): array {
 		$probe->includeNonVisibleCircles()
 			->emulateVisitor();
 
+		if ($invitationCode) {
+			$probe->includeHiddenCircles()
```
artonge: In `LocalController::joinInvitation`, we also call `includeSystemCircles()`, why is it different here?

lib/FederatedItems/CircleCreateInvitation.php
Koc: I've just inspired from already existent `verify` method implementations:

- https://github.com/nextcloud/circles/blob/6c379645ffb836d42a5f0d58fb9c3182ac027afd/lib/FederatedItems/CircleConfig.php#L128
- https://github.com/nextcloud/circles/blob/6c379645ffb836d42a5f0d58fb9c3182ac027afd/lib/FederatedItems/CircleSettings.php#L44
- ...

lib/Model/CircleInvitation.php
Koc: well, I've just inspired from an already existent codebase. Replaced with real types

lib/Db/CircleInvitationRequest.php
Koc: not now. In theory we can add that, but in this case we need rework UI and introduce something like "invitation link name" to distinguish links.

I would like to start with something really simple and improve it later

lib/Db/CircleInvitationRequest.php
Koc: inlined

lib/Db/CircleRequest.php
Koc: no, because we need to respond with `invitationCode` in `GET https://nextcloud.test/ocs/v2.php/apps/circles/circles` endpoint

image

lib/Service/CircleService.php
Koc: added

lib/Db/CircleRequestBuilder.php
Koc: inlined

Koc:
> Error: lib/FederatedItems/CircleCreateInvitation.php:47:36: DeprecatedMethod: The method OCP\Security\ISecureRandom::generate has been marked as deprecated (see https://psalm.dev/001)

> 	 * @deprecated 35.0.0 Use {@see Randomizer::getBytesFromString()} available in PHP 8.3+ instead.

I can't use `Randomizer` as module still relies on [PHP 8.1](https://github.com/nextcloud/circles/blob/master/composer.json#L11)

lib/FederatedItems/CircleJoin.php
```diff
@@ -245,7 +245,8 @@ private function manageMemberStatus(Circle $circle, Member $member) {
 
 			throw new MemberAlreadyExistsException(StatusCode::$CIRCLE_JOIN[122], 122);
 		} catch (MemberNotFoundException) {
-			if (!$circle->isConfig(Circle::CFG_OPEN)) {
+			$allowedToJoin = $invitationCode && $circle->getCircleInvitation()?->getInvitationCode() === $invitationCode;
+			if (!$circle->isConfig(Circle::CFG_OPEN) && !$allowedToJoin) {
```
cristianscheid, line 238: since `invitationCode` is returned in `GET ocs/v2.php/apps/circles/circles`, that would mean every member of the circle, regardless of membership level, can see the code, right? So in the following scenario:

- `circle_abc` does not have `Circle::CFG_OPEN` enabled
- a member of `circle_abc` gets the `invitationCode` from the response (network tab)
- shares the code with john
- john uses it to call `POST ocs/v2.php/apps/circles/invitations/{invitationCode}`

when reaching this if statement, `$allowedToJoin` would resolve to `true`, so even though the circle isn't open, john would be able to bypass the config and join it, no?

lib/Controller/LocalController.php
```diff
@@ -593,6 +593,97 @@ public function link(string $circleId, string $singleId): DataResponse {
 		}
 	}
 
+	#[NoAdminRequired]
+	public function createInvitation(string $circleId): DataResponse {
+		try {
+			$this->setCurrentFederatedUser();
+
+			$outcome = $this->circleService->createInvitation($circleId);
```
cristianscheid, line 610: shouldn't this only be reachable by members with certain levels (e.g. admin, owner)? as it is right now, seems like anyone can create invitations

you can see on other methods of this controller how permissions are handled, for example here:
https://github.com/nextcloud/circles/blob/master/lib/Controller/LocalController.php#L423

lib/Controller/LocalController.php
```diff
@@ -593,6 +593,97 @@ public function link(string $circleId, string $singleId): DataResponse {
...
+			throw new OCSException($e->getMessage(), (int)$e->getCode(), $e);
+		}
+	}
+
+	#[NoAdminRequired]
+	public function revokeInvitation(string $circleId): DataResponse {
+		try {
+			$this->setCurrentFederatedUser();
+
+			$outcome = $this->circleService->revokeInvitation($circleId);
```
cristianscheid, line 631: same as createInvitation()

lib/Controller/LocalController.php
Koc: fixed, but this check in controller duplicates same check inside [`CircleCreateInvitation`](https://github.com/nextcloud/circles/blob/644c2eac0cd81bf308fc2bb4269de3b1c36c8dba/lib/FederatedItems/CircleCreateInvitation.php#L35)

lib/Controller/LocalController.php
Koc: fixed as well

lib/FederatedItems/CircleJoin.php
Koc: > that would mean every member of the circle, regardless of membership level, can see the code, right?
no

image

image

image

Koc:
@artonge @cristianscheid mates, I've resolved/responded to all of your comments. Also this PR contains FE part as well (as we've moving UI for teams management to the Circles module). Please check one more time.

Also I have no idea how to build js to pass corresponding pipeline step, please help here (`npm ci && npm run build` not helps)

miaulalala (approved):
LGTM, some tests would be welcome but not a blocker for me

Koc:
@miaulalala thanx, Anna!

I thought about tests, but:
- Unit tests has no sense here
- Functional tests (Behat) are not bootstrapped in this project at all
- E2e (Cypress/Playwright) tests looks a bit expensive

I'm completely fine to return to that topic and cover new endpoints with Behat tests once we have initial infrastructure for that (like in other modules like Forms/Tables/etc)

AIlkiv (approved):
I did a review — it looks very good.

Issue #2261, opened 2025-12-05
Members (users) not shown correctly in Team


ericfischereu:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

### Steps to reproduce
1. Create a team
2. add members (either internal or guest users)
3. reload the page

### Expected behaviour
All members are listed.

### Actual behaviour
Only the owner, teams and I guess groups are listed.

### Server configuration

Company instance

### Browser

**Browser name:** Zen

**Browser version:** 1.17.12b (Firefox 145.0.2) (64-Bit)

**Operating system:** Linux Mint

<details>
<summary>Browser log</summary>

```
Insert your browser log here, this could for example include:
a) The javascript console log
b) The network log
c) ...
```

</details>

admintechblog:
I have the same issue. 
We had an old owncloud instance named https://system1.domain.ca/. We installed a new Nextcloud instance with version 32.0.3, the Nextcloud instance was originally referred to as https://system2.domain.ca/. We created Teams (circles) and added members. Everything was working fine until our migration. We decommissioned the old Owncloud instance and changed the url for the Nextcloud instance from https://system2.domain.ca/ to https://system1.domain.ca/
After this change the Teams shows initially the number of original members, but if we click on the Teams, the members are not shown and the number of members changes to 1. Also the members are not able to reach their shared resources.

alx-tuilmenau:
@jodergrosse sorry, i used the line numbers from stable31 (because i'm still on NC 31), but linked to master...

mejo-:
Closing this one as it got fixed by #2268.

mejo-:
@rasos, yes this already got backported to Nextcloud 32 and 31. The correct PR is https://github.com/nextcloud/circles/pull/2267 by the way, and there you also see the backport PRs.

mejo-:
@leoniscsem just to be sure, are you on Nextcloud 32.0.4 already?

rasos:
Yes you have to update to a latest minor Nextcloud release. I've learned that fixes in Teams/Circles are not offered as individual app updates. 

[...]

PR #2267, opened 2025-12-18
fix empty memberlist

ArtificialOwl:
Resolves: https://github.com/nextcloud/circles/issues/2261

- regression from a fix to filter circle access from initiator (#2256)
- replacing the previous fix with a better one !


AndyScherzinger (approved):
🎅

PR #2290, opened 2026-01-15
feat(ShareByCircleProvider): Implement IShareProviderGetUsers

provokateurin:


lib/ShareByCircleProvider.php
```diff
@@ -845,4 +850,19 @@ public function getAllShares(): iterable {
 		//		$cursor->closeCursor();
 		return [];
 	}
+
+	public function getUsersForShare(IShare $share): iterable {
+		$members = $this->shareWrapperService->getShareById((int)$share->getId())->getCircle()->getInheritedMembers();
```
ArtificialOwl, line 852: might throw a `ShareWrapperNotFoundException` 
should we manage this ?

lib/ShareByCircleProvider.php
```diff
@@ -845,4 +850,19 @@ public function getAllShares(): iterable {
...
+		foreach ($members as $member) {
+			if ($member->getUserType() === Member::TYPE_USER) {
+				/** @var IUser $user */
+				$user = $this->userManager->get($member->getUserId());
+				yield $user;
+			} elseif ($member->getUserType() === Member::TYPE_GROUP) {
```
ArtificialOwl: member from groups are already returned by `getInheritedMembers`;
I would keep the first condition about `Member::TYPE_USER` but remove the `elseif` block

lib/ShareByCircleProvider.php
provokateurin: That shouldn't happen, because the method should only receive existing shares by the circles provider. I think this exception should never happen in reality, unless there is a race condition, but there is no way to handle that gracefully.

lib/ShareByCircleProvider.php
provokateurin: Ok nice, I didn't know that.

lib/ShareByCircleProvider.php
```diff
@@ -845,4 +847,15 @@ public function getAllShares(): iterable {
...
+
+	public function getUsersForShare(IShare $share): iterable {
+		$members = $this->shareWrapperService->getShareById((int)$share->getId())->getCircle()->getInheritedMembers();
+		foreach ($members as $member) {
+			if ($member->getUserType() === Member::TYPE_USER) {
+				/** @var IUser $user */
+				$user = $this->userManager->get($member->getUserId());
+				yield $user;
+			}
+		}
```
icewind1991, line 859: The group handling doesn't seem to work as expected, when adding a group to a circle, the members of the group aren't returned when calling `getUsersForShare` afterwards.

Adding the user directly to the circle, or adding another circle seem to work as expected

provokateurin: So what @ArtificialOwl said is wrong and the code before was correct?! :thinking:

ArtificialOwl: I did some test with TeamA that has 3 members: 2 users and 1 group, and calling `getInheritedMembers()` returns members from the parent teams, and members from the group

icewind1991: - A fresh instance with user "admin", user "test" and group "test" containing user "test".
- As admin,  create a team "TeamA" and add group "test"
- As admin, share a file with "TeamA"

Placing a breakpoint [inside a `getUserForShare`](https://github.com/nextcloud/server/blob/eff7f8d4f57824a382b66c5014efd8512edb49e6/apps/files_sharing/lib/Listener/SharesUpdatedListener.php#L45) loop, I only see the "admin" user being returned.

icewind1991: There might be something deeper going on here, the "test" user sees neither the created share, nor does "TeamA" show as a team they are part of in their contacts app

ArtificialOwl: can you please run `./occ circles:sync --groups` on your fresh instance ?

icewind1991: I think the issue might be caused by the fact that the group is created while the circles app is disabled, if I create a new group with "test" as member, add it to a circle and share it to the circle then things work as expected.

So it looks like a separate circles bugs around handling pre-existing groups and nothing related to this PR

icewind1991: > can you please run `./occ circles:sync --groups` on your fresh instance ?

Doesn't help

icewind1991: Raised https://github.com/nextcloud/circles/issues/2291 for it

Issue #2126, opened 2025-09-12
Wrong user count when using teams as team owners


dgsiegel:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

### Steps to reproduce
1. Create two teams "foo" and "bar"
2. Add team "foo" to "bar" and set "foo" as sole owner of that team 
3. The sidebar will now show a user count of 2 instead of 1.
4. When visiting the team, the user count reduces to the correct number of 1

### Expected behaviour
Always show the same user count. If you count the team and the single member of "foo" as 2, that would be fine as well.

### Server configuration

**Database:** PostgreSQL

**PHP version:** 8.3

**Nextcloud version:** 31.0.8

PR #2300, opened 2026-01-17
fix(population): exclude nested circles from population counts

cristianscheid:
* Related to: nextcloud/circles#2126

## Summary
Fixes team population calculation to exclude nested teams from member counts.

## Changes
- When a team is added as member of another team, only users within the nested team are counted (team itself and duplicates are not counted)
- When team members are added/removed, all parent circles (circles containing the updated circle) are also updated
- When a team is deleted, all teams that contained that team are updated to reflect the removal
- Added `populationInherited` field to `Circle::jsonSerialize()` alongside existing `population` for frontend usage in follow-up PR without breaking current frontend

### Frontend follow-up needed
This PR addresses only the backend part of the issue. A follow-up PR in `nextcloud/contacts` is required to fix inconsistent count display:
- Currently: uses `population` for member count on page load, but switches to member list length when clicking a team
- Expected: consistently use the same count throughout the interface (`populationInherited`, added in this PR)

## Checklist

### DIFF
--- a/lib/Db/MembershipRequest.php
+++ b/lib/Db/MembershipRequest.php
@@ -13,6 +13,7 @@ namespace OCA\Circles\Db;
 
 use OCA\Circles\Exceptions\MembershipNotFoundException;
 use OCA\Circles\Model\FederatedUser;
+use OCA\Circles\Model\Member;
 use OCA\Circles\Model\Membership;
 use OCP\DB\QueryBuilder\IQueryBuilder;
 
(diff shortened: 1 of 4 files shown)


lib/Db/MembershipRequest.php
```diff
@@ -113,6 +114,36 @@ public function getInherited(string $singleId, int $level = 0): array {
...
+		if ($level > 1) {
+			$qb->andWhere($expr->gte('level', $qb->createNamedParameter($level, IQueryBuilder::PARAM_INT)));
+		}
+		$qb->leftJoin(
+			CoreQueryBuilder::MEMBERSHIPS,
+			CoreRequestBuilder::TABLE_MEMBER,
+			CoreQueryBuilder::MEMBER,
+			$expr->andX(
+				$expr->eq(CoreQueryBuilder::MEMBERSHIPS . '.single_id', CoreQueryBuilder::MEMBER . '.single_id'),
+				$expr->eq(
```
provokateurin, line 137:
```suggestion
				$expr->neq(
```
If the method is supposed to exclude circles?

cristianscheid: Using neq in the JOIN matches each membership with all non-circle members sharing the same single_id, creating multiple rows when a user belongs to multiple circles. The eq + isNull approach avoids this by only joining with circle-type members. Since each single_id can only be a circle once (not multiple times), there's at most one match per membership. The isNull then filters to keep memberships that didn't match any circles, guaranteeing exactly one row per non-circle membership.

PR #2310, opened 2026-01-21
fix(groups): add repair step for group circles sync and adjust 'circles:sync --groups' command

cristianscheid:
Closed accidentally. Reopened in #2317.

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -12,7 +12,7 @@
 Those groups of people can then be used by any other app for sharing purpose.
 		]]>
 	</description>
-	<version>33.0.0-dev.0</version>
+	<version>34.0.0-dev.0</version>
 	<licence>agpl</licence>
 	<author>Maxence Lange</author>
 	<types>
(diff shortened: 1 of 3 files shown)


lib/Service/SyncService.php
```diff
@@ -277,9 +277,13 @@ public function syncNextcloudGroup(string $groupId): Circle {
...
+		$members = [];
+		foreach ($this->memberRequest->getMembers($circle->getSingleId()) as $member) {
+			if ($member->getUserType() === Member::TYPE_APP) {
+				continue;
+			}
+			$members[] = $member->getSingleId();
+		}
 
 		$group = $this->groupManager->get($groupId);
 		if ($group->count() <= count($members)) {
```
provokateurin, line 285: Because you mentioned it in the PR description: I think this check should be removed to avoid sync issues. If the sync is broken for some reason and one user is added and one users is removed, then this would prevent syncing. I suppose this is done for performance reasons, but it might be better to just always sync or at least implement a better detection than the member count.

lib/Service/SyncService.php
```diff
@@ -277,9 +277,13 @@ public function syncNextcloudGroup(string $groupId): Circle {
...
-		$members = array_map(function (Member $member): string {
-			return $member->getSingleId();
-		}, $this->memberRequest->getMembers($circle->getSingleId()));
+		$members = [];
+		foreach ($this->memberRequest->getMembers($circle->getSingleId()) as $member) {
+			if ($member->getUserType() === Member::TYPE_APP) {
+				continue;
+			}
+			$members[] = $member->getSingleId();
+		}
```
provokateurin: IMO array_filter is cleaner for such operations.

PR #2324, opened 2026-01-27
admin settings endpoint

ArtificialOwl:


lib/Controller/SettingsController.php
```diff
@@ -0,0 +1,91 @@
...
+    public function setValue(string $key, string $value): DataResponse {
+        if ($key === 'frontal_cloud') {
+            if ($this->setFrontalValue($value)) {
+                return $this->getValues();
+            }
+
+    		return new DataResponse(['data' => ['message' => 'wrongly formated value']], Http::STATUS_BAD_REQUEST);
+        }
+
+        return $this->getValues();
```
cristianscheid: Currently, seems that invalid keys are silently ignored and return `getValues()`. Maybe we could explicitly handle unsupported keys with an error response instead?

```suggestion
        return new DataResponse(['data' => ['message' => 'unsupported key']], Http::STATUS_BAD_REQUEST);
```

ArtificialOwl: true, I dont mind; my own idea was that we could add keys in the future and I was fine to silently ignore, i'll change this

lib/Controller/SettingsController.php
```diff
@@ -0,0 +1,91 @@
...
+
+	public function setValue(string $key, string $value): DataResponse {
+		if ($key === 'frontal_cloud') {
+			if ($this->setFrontalValue($value)) {
+				return $this->getValues();
+			}
+
+			return new DataResponse(['data' => ['message' => 'wrongly formated value']], Http::STATUS_BAD_REQUEST);
+		}
+
```
cristianscheid, line 37:
```suggestion

		if ($key === 'federated_teams_enabled') {
			$this->appConfig->setAppValueBool('federated_teams_enabled', $value);
			return $this->getValues();
		}

```
Note: `ConfigService::FEDERATED_TEAMS_ENABLED` does not exist yet in `lib/Service/ConfigService.php`

lib/Controller/SettingsController.php
```diff
@@ -0,0 +1,91 @@
...
+		return new DataResponse(['data' => ['message' => 'unsupported key']], Http::STATUS_BAD_REQUEST);
+	}
+
+	public function getValues(): DataResponse {
+		return new DataResponse([
+			'frontal_cloud' => $this->getFrontalValue() ?? '',
```
cristianscheid:
```suggestion
			'frontal_cloud' => $this->getFrontalValue() ?? '',
			'federated_teams_enabled' => $this->appConfig->getAppValueBool('federated_teams_enabled'),
```
Note: `ConfigService::FEDERATED_TEAMS_ENABLED` does not exist yet in `lib/Service/ConfigService.php`

lib/Controller/SettingsController.php
```diff
@@ -0,0 +1,91 @@
...
+		$path = parse_url($url, PHP_URL_PATH);
+
+		if (is_bool($scheme) || is_bool($cloudId) || is_null($scheme) || is_null($cloudId)) {
+			return null;
+		}
+
+		if (is_null($path) || is_bool($path)) {
+			$path = '';
+		}
+		$path = rtrim($path, '/');
```
cristianscheid, line 96: `rtrim()` results in `frontal_cloud_path` being saved as "/some_path" on DB in my tests. Using `trim()` saves it as "some_path"

```suggestion
		$path = trim($path, '/');
```

lib/Controller/SettingsController.php
ArtificialOwl: please use `setAppValueBool()`

lib/Controller/SettingsController.php
ArtificialOwl: please switch to `getAppValueBool()`

lib/Controller/SettingsController.php
ArtificialOwl: i am fine with it

lib/Controller/SettingsController.php
cristianscheid: Suggestion updated accordingly.

lib/Controller/SettingsController.php
cristianscheid: Suggestion updated accordingly.

lib/Controller/SettingsController.php
```diff
@@ -0,0 +1,97 @@
...
+		if ($key === 'frontal_cloud') {
+			if ($this->setFrontalValue($value)) {
+				return $this->getValues();
+			}
+
+			return new DataResponse(['data' => ['message' => 'wrongly formated value']], Http::STATUS_BAD_REQUEST);
+		}
+
+		if ($key === 'federated_teams_enabled') {
+			$this->appConfig->setAppValueBool('federated_teams_enabled', $value);
```
cristianscheid: Cast to bool since setValue() receives string ('yes' -> true, others -> alse)
```suggestion
			$this->appConfig->setAppValueBool('federated_teams_enabled', $value === 'yes');
```

PR #2332, opened 2026-02-02
add trusted servers as external

ArtificialOwl:


lib/Service/TrustedServerService.php
```diff
@@ -0,0 +1,43 @@
...
+
+use OCA\Federation\TrustedServers;
+use Psr\Log\LoggerInterface;
+
+/**
+ * @psalm-suppress UndefinedClass
+ */
+class TrustedServerService {
+	public function __construct(
+		private readonly TrustedServers $trustedServers,
```
nickvergessen, line 19: This throws when the federation app is disabled, breaking all of OCC.

PR #2390, opened 2026-03-12
fix(mail): send internal link

ArtificialOwl:
generate an internal link instead of public link when sending mail to local account


cristianscheid (approved):

T0mWz:
@ArtificialOwl , I wonder if this change is sufficient, as I think that still  a public link token is generated, although this is not actively communicated to the user. https://github.com/nextcloud/circles/blob/master/lib/ShareByCircleProvider.php#L145

This change does not, in fact, resolve the underlying problem. Namely that a shareToken is generated, but it is only not mailed. This means the vulnerability remains, allowing unauthorised access to the data if the correct URL is guessed
```
-		$share->setToken($this->token(15));
+		$share->setToken('');
```

PR #2443, opened 2026-04-15
Bug: Don't generate a public link token for internal circle team share

T0mWz:
**Issue**

When sharing data with a team, a public link token is inadvertently generated alongside the secure internal share link for authenticated users. This public link has the same access rights and, until recently, was shared with team members via email. 
Although the recipients were already authorised to access this data, recipients of the email may have forwarded it to third parties, thereby allowing them – without authorisation – to gain access with the same rights as an authorised person. Team members are no longer informed of this, but a public link token is still generated behind the scenes.
 
This public link token can easily be guessed using a randomiser, thereby allowing unauthorised access to data that should not have been publicly accessible.

**How to reproduce**
Create a team via the contacts. Then share a dataset with custom permissions with this team.

** Result **
```
MariaDB [nextcloud]> select * from oc_share where id = '16135' \G;
*************************** 1. row ***************************
                      id: 16135
              share_type: 7
              share_with: ax1Gvs4s964zpZR5z7kyXgbzd9QPXnh
               uid_owner: piet
           uid_initiator: piet
                  parent: NULL
               item_type: folder
             item_source: 6946132
             item_target: NULL
             file_source: 6946132
             file_target: /Test_Public_Link_Breach
             permissions: 31
                   stime: 1776062454
                accepted: 1
              expiration: NULL
                   token: 8ui9FePpNwznze8
               mail_send: 1
              share_name: NULL
                password:
        password_by_talk: 0
                    note: Bla
           hide_download: 0
                   label: NULL
              attributes: [["permissions","download",true]]
password_expiration_time: NULL
           reminder_sent: 0
1 row in set (0.000 sec)
```
You can view this data without authentication via the public link token (8ui9FePpNwznze8), with the assumption that it has only been shared with team members via authorised access.

### DIFF
--- a/lib/Service/ShareWrapperService.php
+++ b/lib/Service/ShareWrapperService.php
@@ -149,15 +149,13 @@ class ShareWrapperService {
 	 * @throws ShareWrapperNotFoundException
 	 */
 	public function getShareByToken(string $token, ?FederatedUser $federatedUser = null): ShareWrapper {
-        // First try using the standard oc_share.token lookup
-        try {
-            return $this->shareWrapperRequest->getShareByToken($token);
-        } catch (ShareWrapperNotFoundException $e) {
-            // Fallback: search using oc_circles_token.token
-            // This is necessary when oc_share.token is empty or does not match,
-            // but the circles token does exist.
-            return $this->shareWrapperRequest->getShareByCirclesToken($token);
-        }
+		// First try the standard lookup via oc_share.token
+		try {
+			return $this->shareWrapperRequest->getShareByToken($token);
+		} catch (ShareWrapperNotFoundException $e) {
+			// oc_share.token is empty — fall back to oc_circles_token.token
+			return $this->shareWrapperRequest->getShareByCirclesToken($token);
+		}
 	}
 
     /**
--- a/lib/Service/ShareWrapperService.php
+++ b/lib/Service/ShareWrapperService.php
@@ -149,8 +149,27 @@ class ShareWrapperService {
 	 * @throws ShareWrapperNotFoundException
 	 */
 	public function getShareByToken(string $token, ?FederatedUser $federatedUser = null): ShareWrapper {
-		return $this->shareWrapperRequest->getShareByToken($token, $federatedUser);
+        // First try using the standard oc_share.token lookup
+        try {
+            return $this->shareWrapperRequest->getShareByToken($token);
+        } catch (ShareWrapperNotFoundException $e) {
+            // Fallback: search using oc_circles_token.token
+            // This is necessary when oc_share.token is empty or does not match,
+            // but the circles token does exist.
+            return $this->shareWrapperRequest->getShareByCirclesToken($token);
+        }
 	}
+
+    /**
+     * Look up a share directly via oc_circles_token.token.
+     * Can be used as an alternative to getShareByToken when
+     * oc_share.token has been deliberately left empty.
+     *
+     * @throws ShareWrapperNotFoundException
+     */
+    public function getShareByCirclesToken(string $token): ShareWrapper {
+        return $this->shareWrapperRequest->getShareByCirclesToken($token);
+    }
 
 	/**
 	 * @return ShareWrapper[]
--- a/lib/ShareByCircleProvider.php
+++ b/lib/ShareByCircleProvider.php
@@ -142,7 +142,6 @@ class ShareByCircleProvider implements IShareProvider, IPartialShareProvider, IS
 			->add(DataProbe::INITIATOR, [DataProbe::BASED_ON]);
 
 		$circle = $this->circleService->probeCircle($share->getSharedWith(), $circleProbe, $dataProbe);
-		$share->setToken('');
 		$share->setMailSend(true);
 		$owner = $circle->getInitiator();
 		$this->shareWrapperService->save($share);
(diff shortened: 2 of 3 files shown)


T0mWz:
This issue is related to pull request; https://github.com/nextcloud/circles/pull/2390

T0mWz:
@cristianscheid please can you review this?

cristianscheid:
Hi @T0mWz, thanks for the PR! I'm currently reviewing and testing the impact of this fix. I'll keep you posted.

cristianscheid:
After testing the fix locally, I found it introduces a new bug when sharing a folder with a team that has members added by email (like below).
image
When trying to access the link sent to those users, they no longer work as expected. I'm still investigating the impacts of this fix on teams that include federated users.

T0mWz:
Hi @cristianscheid ,

Sorry for my late reply. You’re absolutely right. I hadn’t tested this properly either; I can now see the Internal Server Error on my end too, although it’s still with NC32.

With Nextcloud Circles, a share can be created with users within Nextcloud. However, external users can also be added based on their email address.

When sharing via a Nextcloud Circle, a share type 7 is added to the oc_share table. Separate from the share with the circle, a public link token is also added. This token is now directly accessible to everyone without any form of security, but is necessary for the underlying circle token.

This patch omits the setToken, but it is still required behind the scenes for those receiving the share via the Circle.

Currently, public links rely entirely on the oc_share token entry. With my recently (extra) pushed patch, a change has been added to the getShareByToken function in the circle app, ensuring that a circles-token lookup also takes place in oc_circles_token and is not dependent on oc_share.token. The oc_share.token can then remain empty or be assigned a non-routable value.

Could you check whether this solves the problem for you as well?

Thanks, Tom

lib/ShareByCircleProvider.php
```diff
@@ -142,7 +142,7 @@ public function create(IShare $share): IShare {
 			->add(DataProbe::INITIATOR, [DataProbe::BASED_ON]);
 
 		$circle = $this->circleService->probeCircle($share->getSharedWith(), $circleProbe, $dataProbe);
-		$share->setToken($this->token(15));
+		$share->setToken('');
```
artonge: Do we need to set an empty string? Can we instead drop that line?

lib/ShareByCircleProvider.php
```diff
@@ -581,13 +581,32 @@ public function getShareByToken($token): IShare {
...
+        try {
+            $shareWrapper = $this->shareWrapperService->getShareByToken($token);
+        } catch (ShareWrapperNotFoundException $e) {
+            // Not found via oc_share.token — try oc_circles_token.token
+            try {
+                $shareWrapper = $this->shareWrapperService->getShareByCirclesToken($token);
+            } catch (ShareWrapperNotFoundException $e2) {
+                throw new ShareNotFound(
+                    'Share not found for token (tried both oc_share and oc_circles_token): ' . $token
+                );
```
artonge: Looks like a duplicate of the fallback in `getShareByToken`, no?

lib/Service/ShareWrapperService.php
```diff
@@ -149,9 +149,28 @@ public function getSharesByFileIds(array $fileIds, bool $getData = false, bool $
...
 	 */
 	public function getShareByToken(string $token, ?FederatedUser $federatedUser = null): ShareWrapper {
-		return $this->shareWrapperRequest->getShareByToken($token, $federatedUser);
+        // First try using the standard oc_share.token lookup
+        try {
+            return $this->shareWrapperRequest->getShareByToken($token);
+        } catch (ShareWrapperNotFoundException $e) {
+            // Fallback: search using oc_circles_token.token
+            // This is necessary when oc_share.token is empty or does not match,
+            // but the circles token does exist.
```
artonge: I don't think that is true, I think `getShareByToken` will match the circle token, meaning that the fallback is useless.

lib/ShareByCircleProvider.php
```diff
@@ -581,13 +581,32 @@ public function getShareByToken($token): IShare {
...
+
+        if ($shareWrapper === null) {
+            throw new ShareNotFound('Share not found for token: ' . $token);
+        }
+
+        // Ensure that the share object has the correct token for UI routing
+        // If oc_share.token is empty, use the provided circles token
+        if ($shareWrapper->getToken() === '' || $shareWrapper->getToken() === null) {
+            $shareWrapper->setToken($token);
+        }
```
artonge: Maybe we should simply return the `oc_circles_token.token` in `ShareWrapper->getToken()` instead of the `oc_share.token`?

lib/ShareByCircleProvider.php
```diff
@@ -581,13 +581,32 @@ public function getShareByToken($token): IShare {
...
+
+		$share = $shareWrapper->getShare($this->rootFolder, $this->userManager, $this->urlGenerator);
```
artonge, line 589: Why do we need this?

artonge (changes requested):

lib/ShareByCircleProvider.php
T0mWz: You're right. I've just tested it. If I don't set this, it becomes NULL in the database. I'll remove the line

lib/ShareByCircleProvider.php
T0mWz: We have to convert the  internal ShareWrapper object from the circles app, the database output, into a Nextcloud IShare object.

T0mWz:
Thanks for your feedback @artonge !

**What I tested:**
I’ve made a few improvements and also double-checked that it works with:
 1. A default oc_share.token which is used for public link shares
 2. A circles_token (rather than a public link token)
 3. An invalid token that doesn’t match either an oc_share.token or a circles_token

**What changed:**
- ShareWrapperRequest: add getShareByCirclesToken() which joins  oc_circles_token to find a share by its circles token
- ShareWrapperService: add fallback to getShareByCirclesToken()  when getShareByToken() throws ShareWrapperNotFoundException
- ShareByCircleProvider: set the circles token on the IShare object  after construction if oc_share.token is empty

**Summary of this change:**
When sharing with a Nextcloud Circle that includes external members, a public link token is stored in oc_share.token alongside the actual circles token in oc_circles_token.token. This public link is accessible to anyone without authentication, which is undesirable.

To prevent unauthenticated access, oc_share.token can be left empty. 

This change adds a fallback in ShareWrapperService::getShareByToken() that looks up the share via oc_circles_token.token when no match is found in oc_share.token. After the IShare object is constructed from the  database, the circles token is applied to it directly, ensuring the router always receives a valid non-empty token.

cristianscheid:
The issue this PR aimed to address should now be resolved by this other PR: https://github.com/nextcloud/circles/pull/2454. Shall we go ahead and close this one?

T0mWz:
Should be fixed with MR; https://github.com/nextcloud/circles/pull/2454

PR #2449, opened 2026-04-22
fix(shares): Team shares are reached via membership mount only

nfebe:
Resolving a Team share by its token now returns not-found. Members access Team shares through the mount they receive from their Team membership, never through the public share token route. The token field is still populated on the share row because it is reused as an internal key for federated Team mounts.


cristianscheid:
Closing this PR since it's been a while since it was opened, with no activity since then.

PR #2544, opened 2026-06-19
test(admin-controller): add unit tests covering all controller endpoints

cristianscheid:
* Resolves: #

## Summary

Refactor `testCirclesList()` and add coverage for all remaining endpoints in `lib/Controller/AdminController`.

## Checklist

- [x] Tests ([unit](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#unit-tests), [integration](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#integration-tests), api and/or acceptance) are included

## AI (if applicable)

### DIFF
--- a/tests/unit/lib/Controller/AdminControllerTest.php
+++ b/tests/unit/lib/Controller/AdminControllerTest.php
@@ -1,7 +1,7 @@
 <?php
 
 /**
- * SPDX-FileCopyrightText: 2022 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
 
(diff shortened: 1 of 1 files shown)


tests/unit/lib/Controller/AdminControllerTest.php
```diff
@@ -1,118 +1,394 @@
...
 
-	/** @var FederatedUserService|MockObject */
-	private $federatedUserService;
+		$app = new Application();
+		$c = $app->getContainer();
 
-	/** @var CircleService|MockObject */
-	private $circleService;
+		foreach ([self::TEST_USER_1, self::TEST_USER_2] as $userId) {
+			$user = $c->get(IUserManager::class)->get($userId);
```
come-nc: You should get the user manager only once and store it in a var

tests/unit/lib/Controller/AdminControllerTest.php
```diff
@@ -1,118 +1,394 @@
...
+		$this->assertSame($result['name'], 'test-circle');
+		$this->assertSame($result['population'], 0);
+		$this->assertSame($result['config'], 0);
+		$this->assertSame($result['initiator']['userId'], self::TEST_USER_1);
+		$this->assertSame($result['initiator']['level'], Member::LEVEL_OWNER);
+		$this->assertSame($result['owner']['userId'], self::TEST_USER_1);
+	}
+
+	public function testDestroy(): void {
+		$c = $this->container;
```
come-nc:
```suggestion
		$circleService = $this->container->get(CircleService::class);
```

come-nc: Same in the other tests

come-nc (commented):
I think it’s the right kind of tests.
You could use assertSame directly on the array instead of each key/value I think.

cristianscheid:
> I think it’s the right kind of tests. You could use assertSame directly on the array instead of each key/value I think.

@come-nc thank you very much for the review.

The reason I didn't use `assertSame` on the entire array on some tests like `$this->assertSame($result, $circleData)` is because the structure returned from `CircleService` methods is slightly different from the one returned by the `AdminController` methods. So instead, I compared what I judged the more relevant key/values for each scenario.

Your other suggestions were addressed on [commit 5a6eef8](https://github.com/nextcloud/circles/pull/2544/changes/5a6eef840f0649e6f2e3c9e44379aaa94eee3437)

come-nc (approved):

PR #2561, opened 2026-06-26
feat(teams): in-app Teams page

### LINKED ISSUE #2543: Own frontend and app icon for TeamPage
### Use case
The Teams feature is becoming increasingly important, especially for larger organizations. Currently it’s buried too deep inside the Contacts app, and the lack of features limits the app’s huge potential.

The **TeamPage**, as the front end for a team, could become the daily starting point for employees and an aggregation of team-related information, such as:

- Description
- Recently added files
- Activity in decks
- Calendar events
- Collective pages
- Tree view of the files in a linked team-folder
- Buttons to add those ressources

### Proposal
Give Teams (circles) its own frontend and app icon.
This feature request is thought to be an overview issue, for a multitude of features that will be linked to this structural change of the circles app.

marcoambrosini:
Teams app front-end first step: mostly moved the functionality from contacts and added an overview page.
Screenshot 2026-06-26 at 20 08 13
Screenshot 2026-06-26 at 20 08 19


susnux:
So no longer part of contacts app? Will simplify things :)

jancborchardt (approved):
Really really nice! :) Great first step in the right direction, and keeping it simple & contained by just porting over what we have.

jancborchardt:
Small details but no blocker @marcoambrosini :
- [ ] Search input should probably go even above "Add", so that we have a common position at least in height for the in-app search functions.
- [x] The "Overview" icon in the nav should be filled when active?
- [ ] The avatars of the teams in the nav seem a bit too large, there is no space towards the text

marcoambrosini:
@susnux could you have a look at the front-end tooling and wiring and tell me if it's sane?

marcoambrosini:
> Search input should probably go even above "Add", so that we have a common position at least in height for the in-app search functions.

@jancborchardt then I would do it same as talk here, with the new button inline

lib/Controller/PageController.php
susnux, line 1: This should use Route annotation instead of legacy `routes.php` - `routes.php` is only for legacy routes as this is a new file it directly should use the annotations

src/teams/components/TeamCard.vue
```diff
@@ -0,0 +1,181 @@
...
+
+			<span class="team-card__resources">
+				<NcIconSvgWrapper :path="mdiFolderMultipleOutline" :size="18" inline />
+				{{ t('circles', '{count} resources', { count: team.resources.length }) }}
+			</span>
+		</div>
+	</RouterLink>
+</template>
+
+<style scoped lang="scss">
```
susnux: I would use `style module` for new files as this is the only option to 2-way scope styles properly and not leak.

src/teams/team-page/components/CircleDetails/CirclePasswordSettings.vue
```diff
@@ -0,0 +1,263 @@
...
+					<div v-if="uniquePasswordError" class="unique-password-error">
+						{{ t('contacts', 'Failed to save password. Please try again later.') }}
+					</div>
+				</li>
+			</ul>
+		</li>
+	</ul>
+</template>
+
+<script>
```
susnux: This should use `script setup lang="ts"` as all other new files.

src/teams/team-page/components/CircleDetails/CirclePasswordSettings.vue
```diff
@@ -0,0 +1,263 @@
...
+// @ts-nocheck
```
susnux: why?

src/teams/team-page/components/CircleDetails/CirclePasswordSettings.vue
```diff
@@ -0,0 +1,263 @@
...
+import {
+	NcCheckboxRadioSwitch as CheckboxRadioSwitch,
+	NcButton,
+} from '@nextcloud/vue'
```
susnux: You should not import component from main entry point


```suggestion
import NcCheckboxRadioSwitch from '@nextcloud/vue/components/NcCheckboxRadioSwitch'
import NcButton from '@nextcloud/vue/components/NcButton'
```

src/teams/team-page/components/CircleDetails/CircleSettings.vue
```diff
@@ -0,0 +1,163 @@
...
+			<template #icon>
+				<IconDelete :size="20" />
+			</template>
+			{{ t('contacts', 'Delete team') }}
+		</NcButton>
+	</div>
+</template>
+
+<script lang="ts">
+// @ts-nocheck
```
susnux: Same here new files should directly use `script setup lang="ts"`

src/teams/team-page/components/CircleDetails/CircleSettings.vue
```diff
@@ -0,0 +1,163 @@
...
+		<!-- delete circle -->
+		<NcButton
+			v-if="circle.canDelete"
+			variant="error"
+			href="#"
+			@click.prevent.stop="$emit('delete')">
+			<template #icon>
+				<IconDelete :size="20" />
+			</template>
+			{{ t('contacts', 'Delete team') }}
```
susnux: wrong app used here and in other places

Koc:
hey-hey!

Looks really promising. @marcoambrosini Will it work fine with large circles (1000+ members)? See my other PRs related to that https://github.com/nextcloud/collectives/pull/2518, https://github.com/nextcloud/contacts/pull/5330, so maybe you can backport some stuff from them

src/teams/components/CreateTeamDialog.vue
```diff
@@ -0,0 +1,95 @@
...
+		return
+	}
+	submitting.value = true
+	try {
+		const team = await store.createTeam(name.value, description.value)
+		showSuccess(t('circles', 'Team "{name}" created', { name: name.value.trim() }))
+		if (team) {
+			router.push({ name: 'team', params: { teamId: team.id } })
+		}
+		open.value = false
```
pringelmann: Found a bug here: open the dialog, hit "Cancel", then try "New team" again and nothing happens. Same after a successful create. Both close by flipping the local `open` ref, but the parent only unmounts on `@close`, which comes from `onUpdateOpen` and that only fires when NcDialog closes itself (Esc/backdrop/X). So `createDialogOpen` stays true and you can't reopen. Can we just emit `close` from submit and Cancel instead?

lib/Controller/PageController.php
```diff
@@ -0,0 +1,51 @@
...
+	public function __construct(
+		IRequest $request,
+	) {
+		parent::__construct(Application::APP_ID, $request);
+	}
+
+	#[NoAdminRequired]
+	#[NoCSRFRequired]
+	#[FrontpageRoute(verb: 'GET', url: '/teams')]
+	public function index(): TemplateResponse {
```
pringelmann: The dashboard widget checks `frontend_enabled` but this route and the nav entry don't, so Teams shows up even when the frontend is turned off. Intended?

src/teams/components/TeamCard.vue
```diff
@@ -0,0 +1,181 @@
...
+						hide-status />
+				</li>
+				<li v-if="team.memberCount > team.members.length" :class="$style.teamCardMemberMore">
+					+{{ team.memberCount - team.members.length }}
+				</li>
+			</ul>
+
+			<span :class="$style.teamCardResources">
+				<NcIconSvgWrapper :path="mdiFolderMultipleOutline" :size="18" inline />
+				{{ t('circles', '{count} resources', { count: team.resources.length }) }}
```
pringelmann: This should be `n()`, otherwise a team with one resource reads "1 resources"

src/teams/api.ts
```diff
@@ -0,0 +1,215 @@
...
+}
+
+/**
+ * Fetch all of the current user's teams, merging team metadata (name,
+ * description, member count, our role) with the members preview and resources.
+ */
+export async function fetchTeams(): Promise<Team[]> {
+	const [circlesRes, dashRes] = await Promise.all([
+		axios.get<OcsResponse<RawCircle[]>>(generateOcsUrl('apps/circles/circles') + '?limit=-1', { headers: HEADERS }),
+		axios.get<OcsResponse<RawDashboardTeam[]>>(generateOcsUrl('apps/circles/teams/dashboard/widget') + '?limit=200&offset=0', { headers: HEADERS }),
```
pringelmann, line 131: If the dashboard call fails but `/circles` succeeds, `Promise.all` rejects and the whole page shows the error state even though we have the team list. `allSettled` here and treat missing dashboard data as empty previews maybe?

pringelmann (changes requested):
Looks great! 

~If you rebase on master you'll find new testing configs for FE unit and E2E tests. If possible could you please add a few critical/happy path tests?~  

EDIT: tests will calcify the code too much this early in the process. Let's defer

stediefan (approved):
Looks great! I already built some UI stuff ontop - can't wait to publish after this gets merged. One further task would be to remove the corresponding code from contacts - right?

marcoambrosini:
Hi @Koc, yes we'll have to handle those too thanks for sharing!

src/teams/components/CreateTeamDialog.vue
marcoambrosini: Wow nice catch!

lib/Controller/PageController.php
marcoambrosini: Not intended, fixed

pringelmann (approved):

PR #2568, opened 2026-07-01
Avoid `occ circles:check` from failing because of outdated cache values

salmart-dev:
## Summary

The `occ circles:check` command sets and reads values from `IAppConfig`, and so does the service called by it, to perform checks.

During the loopback check, the code sets the value in the app config for a dummy token, before issuing a request that reads this value. Since the app config can use caches, when the local cache is set to APCu, the cache is not shared between CLI and HTTP workers. This results in random failures caused by the fact that a worker may receive the HTTP request and still hold the out-of-date config value.

## Checklist

## AI (if applicable)

### DIFF
--- a/lib/Command/CirclesCheck.php
+++ b/lib/Command/CirclesCheck.php
@@ -10,6 +10,7 @@ namespace OCA\Circles\Command;
 
 use Exception;
 use OC\Core\Command\Base;
+use OC\Memcache\APCu;
 use OCA\Circles\AppInfo\Application;
 use OCA\Circles\AppInfo\Capabilities;
 use OCA\Circles\Exceptions\FederatedEventException;
@@ -39,6 +40,7 @@ use OCA\Circles\Tools\Traits\TArrayTools;
 use OCA\Circles\Tools\Traits\TNCRequest;
 use OCA\Circles\Tools\Traits\TStringTools;
 use OCP\IAppConfig;
+use OCP\IConfig;
 use Symfony\Component\Console\Helper\QuestionHelper;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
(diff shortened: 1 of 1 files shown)


salmart-dev:
Completes: https://github.com/nextcloud/circles/pull/2387 as the issue was more about the timing of the caching, than the order with which the config is set.

CarlSchwan (approved):

lib/Command/CirclesCheck.php
```diff
@@ -262,6 +265,17 @@ private function testLoopback(InputInterface $input, OutputInterface $output): b
...
 
+		$localIsApcu = ltrim($this->config->getSystemValueString('memcache.local'), '\\') === '\OC\Memcache\APCu';
+		if ($localIsApcu) {
+			/*
+			 * APCu cache is not shared between the CLI process and web requests, so freshly
+			 * set values will either be missing or stale and make the check fail.
+			 * The amount of seconds slept here depends on the TTL set by the AppConfig.
+			 */
+			$output->writeln('- Waiting for APCu config cache to refresh (4s)');
+			sleep(4);
```
cristianscheid, line 276: is there a way we could get the value (4) from somewhere instead of hardcoding it?

salmart-dev: Besides putting it into a const in this PR, right now no, as the value is in a private const in `AppConfig` (the private implementation of `IAppConfig`).

An alternative idea would have been to be able (via AppConfig) to reset the cache for only one item, then modify the dummy token to include a variable part, and check for that in the app config, if the values don't match, reset the cache for that one item and then check again. But right now one can only reset the entire cache, which is really bad, since if we'd do it when that dummy token string is received, this could open the door to a DoS attack by repeatedly causing caches to be reset.

cristianscheid (approved):

PR #2735, opened 2026-08-10
fix(teams): hide dot entries and route team space link

Fin-c:
* Resolves: # 

## Summary
Hiding .apps Folder in TeamFolder Issue #46309

## TODO

- [ ] ...

## Checklist

## AI (if applicable)

### DIFF
--- a/src/teams/components/TeamFolderWidget.vue
+++ b/src/teams/components/TeamFolderWidget.vue
@@ -98,6 +98,16 @@ function encodeDir(dir: string): string {
 }
 
 /**
+ * Hidden entries in Nextcloud start with a dot and should not be shown in
+ * the Team space widget.
+ *
+ * @param node - The team folder node
+ */
+function isHiddenEntry(node: INode): boolean {
+	return node.basename.startsWith('.')
+}
+
+/**
  * Build the URL for a single node so the list items are real links.
  *
  * Both files and folders open in the Files app using the file id based URL
@@ -222,6 +232,7 @@ async function loadContents(): Promise<void> {
 		nodes.value = data
 			.slice(1)
 			.map((entry) => resultToNode(entry, rootPath))
+			.filter((node) => !isHiddenEntry(node))
 			.sort((a, b) => {
 				if (a.type === b.type) {
 					return a.basename.localeCompare(b.basename)


alimmroth:
As it was agreed on by the Nextcloud team, please rename the folder to `.system`

src/teams/team-page/components/CircleDetails.vue
```diff
@@ -941,6 +941,35 @@ export default {
 			console.debug('Team resources', this.resources)
 		},
 
+		isTeamSpaceResource(resource) {
+			const providerId = (resource?.provider?.id || '').toLowerCase()
+			if (providerId === 'team-folders' || providerId === 'teamfolders' || providerId === 'teamfolder') {
```
artonge: Why so many possibilities?

artonge (commented):
At some point it will make sense to reuse the files app UI through a plugin system instead of re-implementing all the features here

src/teams/team-page/components/CircleDetails.vue
Fin-c: you are right. i will fix that

Fin-c: done!

src/teams/team-page/components/CircleDetails.vue
artonge, line 1: What is this change adding?

Fin-c: This part replaces direct resource URLs with context-aware link resolution. Team Space resources are detected (provider/id/mountpoint) and routed to the internal team route, while all non-Team-Space resources still use their original resource URL.

artonge: Can you provide an example because I am struggeling to hunderstand it.
Am I correct that the branch where it is a team ressource does not use the ressource iteself and returns a generic url?

Fin-c: You are correct: in the team-space branch we intentionally do not use resource.url. We always resolve to the generic team route so team spaces open through the team view instead of provider-specific links. For non-team-space resources, we still use resource.url as before.

It's especially important to me that the “.folder” entry disappears from the resource list so that users don't see a duplicate or misleading entry. I can leave the routing out for now. In my tests, the Teams Space routing in the Manage Teams view caused only one error, and I wanted to fix that.

artonge: Alright, I get it now. Shouldn't we instead change the URL of the teamfolder ressource? Unsure where this is generated and whether that makes sense though.

Fin-c: I think that makes sense, and I agree that this should be fixed where the Team Space resource URL is generated (Groupfolders), rather than in the Circles user interface. In that case, I would remove it here for now. I'll see if that works.

artonge (approved):

PR #2739, opened 2026-08-11
fix(well-known): fallback on manual extraction of params

ArtificialOwl:
In some local network/internal/globalscale/test setup, get url params might not be extracted and a manual operation might be needed.

### DIFF
--- a/lib/Handlers/WebfingerHandler.php
+++ b/lib/Handlers/WebfingerHandler.php
@@ -44,11 +44,21 @@ class WebfingerHandler implements IHandler {
 
 		$request = $context->getHttpRequest();
 		$subject = $request->getParam('resource', '');
+		// In some local network use cases, params might not be available in the request.
+		// We extract them manually and feed the $fallback array
+		if ($subject === '') {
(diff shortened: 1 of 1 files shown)


lib/Handlers/WebfingerHandler.php
```diff
@@ -44,19 +44,29 @@ public function handle(string $service, IRequestContext $context, ?IResponse $pr
 
 		$request = $context->getHttpRequest();
 		$subject = $request->getParam('resource', '');
+		// In some local network use cases, params might not be available in the request.
+		// We extract them manually and feed the $fallback array
+		if ($subject === '') {
+			$params = substr($request->getRequestUri(), strlen($request->getPathInfo()));
+			if (str_starts_with($params, '?')) {
+				parse_str(substr($params, 1), $fallback);
+				$subject = $fallback['resource'];
```
cristianscheid:
```suggestion
				$subject = $fallback['resource'] ?? '';
```

cristianscheid (approved):
only minor suggestion, other than that LGTM

PR #2742, opened 2026-08-11
feat: show activities for teams

### LINKED ISSUE #2692: Show activity of team in Teams app
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

## Feature request

Would be good to be able to follow activity of a team in the Teams app, seeing work on files, Deck boards, chat messages etcetera.

**Is your feature request related to a problem? Please describe.**
It can be hard to track what is happening in your team!

**Describe the solution you'd like**
When you visit the Team app, there is a widget or such that shows an activity feed of your team.

Which activity should it show exactly?
* [x] anything that happens to data owned by the team
* [x] anything that happens to data shared with the team

**Describe alternatives you've considered**
Perhaps in addition you can filter in the Activity app by team. But the Teams app is the most logical place to see this.

**Additional context**
* Related ticket in the Activity app: https://github.com/nextcloud/activity/issues/2666
* This should cover what the creator of issue 75 wanted: https://github.com/nextcloud/circles/issues/75
* we should also track activity of federated team members!

hiobjan:
Co-authored-by :GitHub Copilot (Claude Sonnet 4.6) <copilot@github.com>

* Resolves: #2692 

## Summary
- Fixes the creation of team events: Adding, editing and deleting a team, adding or removing a member and editing the member level adds an activity entry.
- The activities for a team are shown inside the teams app

Before: 
image

After:

New Action
image
New view opened after clicking on the new "Team activity" action
image

## TODO

- [ ] Show activities for team owned resources

## Checklist

## AI (if applicable)

### DIFF
--- a/lib/Activity/ProviderParser.php
+++ b/lib/Activity/ProviderParser.php
@@ -225,11 +225,13 @@ class ProviderParser {
 	 * @return array<string,string|integer>
 	 */
 	protected function generateCircleParameter(array $circle): array {
+		$quotedName = '"' . $circle['name'] . '"';
+
 		return [
 			'type' => 'circle',
 			'id' => $circle['singleId'],
-			'name' => $circle['name'],
-			'_parsed' => $circle['name'],
+			'name' => $quotedName,
+			'_parsed' => $quotedName,
 			'link' => $this->url->getAbsoluteURL($circle['url']),
 		];
 	}
--- a/lib/Db/CoreRequestBuilder.php
+++ b/lib/Db/CoreRequestBuilder.php
@@ -51,6 +51,7 @@ class CoreRequestBuilder {
 	/** @var array */
 	public static $tables = [
 		self::TABLE_CIRCLE => [
+			'id',
 			'unique_id',
 			'name',
 			'display_name',
--- a/lib/Model/Circle.php
+++ b/lib/Model/Circle.php
@@ -156,6 +156,9 @@ class Circle extends ManagedModel implements IEntity, IDeserializable, IQueryRow
 	private $singleId = '';
 
 	/** @var int */
+	private $circleId = 0;
+
+	/** @var int */
 	private $config = 0;
 
 	/** @var string */
@@ -243,6 +246,24 @@ class Circle extends ManagedModel implements IEntity, IDeserializable, IQueryRow
 	 */
 	public function getSingleId(): string {
 		return $this->singleId;
+	}
+
+	/**
+	 * @param int $circleId
+	 *
+	 * @return self
+	 */
+	public function setCircleId(int $circleId): self {
+		$this->circleId = $circleId;
+
+		return $this;
+	}
+
+	/**
+	 * @return int
+	 */
+	public function getCircleId(): int {
+		return $this->circleId;
 	}
 
 	/**
@@ -820,6 +841,7 @@ class Circle extends ManagedModel implements IEntity, IDeserializable, IQueryRow
 	 */
 	public function jsonSerialize(): array {
 		$arr = [
+			'circleId' => $this->getCircleId(),
 			'id' => $this->getSingleId(),
 			'name' => $this->getName(),
 			'displayName' => $this->getDisplayName(),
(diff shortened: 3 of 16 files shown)


lib/Activity/ProviderParser.php
artonge, line 1: Why is this needed?

lib/Model/Circle.php
```diff
@@ -874,7 +896,8 @@ public function importFromDatabase(array $data, string $prefix = ''): IQueryRow
 			throw new CircleNotFoundException();
 		}
 
-		$this->setSingleId($this->get($prefix . 'unique_id', $data))
+		$this->setCircleId($this->getInt($prefix . 'id', $data))
```
artonge, line 901: I don't think that the `id` of a circle be used. At least, I never seen it used. cc. @cristianscheid or @ArtificialOwl for confirmation that this is ok

lib/Db/CoreRequestBuilder.php
```diff
@@ -51,6 +51,7 @@ class CoreRequestBuilder {
 	/** @var array */
 	public static $tables = [
 		self::TABLE_CIRCLE => [
+			'id',
```
cristianscheid, line 54: As @artonge mentioned in his review, I don't see a reason to use the `id` column from a circle's entry in the `oc_circles_circle` table to reference a circle. The unique identifier we use across the codebase is the circle's `singleId`, which corresponds to the `unique_id` column in the table. In some places you'll see this referred to as `singleId` or `circleId` in variable names.

You can get/set a circle's identifier using the existing methods `lib/Model/Circle` (`getSingleId` and `setSingleId`).

lib/Model/Circle.php
```diff
@@ -155,6 +155,9 @@ class Circle extends ManagedModel implements IEntity, IDeserializable, IQueryRow
 	/** @var string */
 	private $singleId = '';
 
+	/** @var int */
+	private $circleId = 0;
```
cristianscheid, line 159: As mentioned in the previous comment, no need to add the circle's `id` here, should use `singleId` instead.

lib/Model/Circle.php
cristianscheid: Also don't think it should be used, left a comment on my review about it as well.

lib/Db/CoreRequestBuilder.php
hiobjan: We currently cannot use the `singleId` because it is a string. The Activity App expects an integer for the `objectId`. I agree that two ids are confusing.
We either use the `id`, or we have to support objectIds of type string in the activity app.

cristianscheid: Another scenario that comes to mind is a federated environment (e.g. instance A + instance B). `singleId` is what we use to identify a circle across multiple instances, so instance A and instance B know that a circle with `singleId = xyz` refers to the same circle.

If we used `id` instead, instance A would trigger an event using `id` and broadcast it to instance B. Instance B would then try to identify the circle using that same `id`, but there's no guarantee that it refers to the same circle (`singleId`) as on instance A, which would cause a mismatch.

jospoortvliet:
So does this this show activity of team members on team-shared or team-owned resources, or only activities on the team itself (adding/removing members etc) or all activities by team members regardless of it being on shared resources?

lib/Model/Circle.php
artonge: After discussing it with @come-nc, we think it make sense to make it so that activity support string ids. We'll need to discuss how to make that happen though. But timeline wise, this probably won't be in 35 :/

hiobjan:
> So does this this show activity of team members on team-shared or team-owned resources, or only activities on the team itself (adding/removing members etc) or all activities by team members regardless of it being on shared resources?

It shows only the activities of the team itself, but could be extended to events of the team resources.

lib/Db/CoreRequestBuilder.php
hiobjan: Decision: we enable the actiity app to use strings as objectId. Because aof this I close this PR

artonge: After some discussion with the team:

1. We won't change the type of the activity table
2. The APIs of activity should accept the string ID of Circles, and do a mapping from the Circles's singleId to its internal ID. Either in a hacky and hardcoded way, or by creating a new API that only Circles would use for now.

Issue #2717, opened 2026-08-04
Allow restricting team-creation to certain groups


alimmroth:
## Proposal
Create an admin setting to restrict team-creation to certain groups of users.

## Use Case
On large installations, the creation of teams was restricted for organisational reasons through restricting the contacts app to certain groups. This no longer works after Teams gets its own frontend.

Also this options is finer grained, still allowing users to actually use and see teams, without being abel to create them.

jospoortvliet:
> ## Use Case
> 
> On large installations, the creation of teams was restricted for organisational reasons through restricting the contacts app to certain groups. This no longer works after Teams gets its own frontend.
> 
> Also this options is finer grained, still allowing users to actually use and see teams, without being abel to create them.

Could we come up with a use case/argument for this that actually hinges on any customer/user/organizational benefit of restricting the creation of teams?

I can't really think of any, to be honest. Teams were meant to be user-managed, and helpful for people to organize their work. This is a bit like disabling the creation of folders in Files, or even disabling sharing in Files. You take away a way for people to organize their work, we better have a good reason to do this.

I worry that admins think that Teams are similar to Teams in Microsoft Teams, that is, that they are very impactful and costly on the organization. But this is not (meant to be) the case - teams can be entirely private and have no impact on the organization.

If we can come up with some actual use cases where you might want to restrict team creation. maybe we can come up with more targeted restrictions.

For example, if we are worried that teams come with storage use, restrict the storage usage in a smarter way.

If we are worried teams become messy, we could restrict the creation if teams that are visible to others, or make it so that teams created by someone in one group can only be seen by others in the same group, for example.

I find 'disabling team creation' going against the concept of Teams.

PR #2698, opened 2026-07-31
feat: add Teams admin settings and group-based creation limits

lennart441:
* Resolves: https://github.com/nextcloud/circles/issues/2717 

## Summary
Adds a Teams admin settings section where admins can restrict which Nextcloud groups may create teams. The restriction is enforced in the API and UI: create actions stay hidden without permission, and the empty-state copy no longer prompts users to create a team.

## Checklist

- Code is [properly formatted](https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/continuous_integration.html#linting)
- [Sign-off message](https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md) is added to all commits
- [x] Tests ([unit](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#unit-tests), [integration](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#integration-tests), api and/or acceptance) are included
- [x] Screenshots before/after for front-end changes
- [x] Documentation ([manuals](https://github.com/nextcloud/documentation/) or wiki) has been updated or is not required
- [x] [Backports requested](https://github.com/nextcloud/backportbot/#usage) where applicable (ex: critical bugfixes)
  no
- [ ] [Labels added](https://github.com/nextcloud/server/labels) where applicable (ex: bug/enhancement, `3. to review`, feature component)
- [ ] [Milestone added](https://github.com/nextcloud/server/milestones) for target branch/version (ex: 32.x for `stable32`)

## AI (if applicable)

- [x] The content of this PR was partly or fully generated using AI
Assisted by: Cursor:Grok-4.5

Screenshot of the teams app without permission to create a team
Bildschirmfoto_20260731_130547

Admin settings of teams after
Bildschirmfoto_20260731_130402
]

PR #2743, opened 2026-08-11
feat: control team space provisioning via app config

lennart441:
feat: kill switch for user team-folder provisioning (app config, no admin UI)

Assisted by: Cursor

Related:
- Supersedes / follows up on https://github.com/nextcloud/circles/pull/2698
- Issue from that PR: https://github.com/nextcloud/circles/issues/2717
- Review direction: keep teams creatable; restriction via CLI / app config, **not** admin UI

## Summary
Kill switch for **user** team-folder provisioning so admins keep quota control. Teams stay creatable.

When disabled:
- no auto-create on team creation
- no Circles UI/API self-upgrade

```bash
occ config:app:set circles team_folder_auto_create --value="false" --type=boolean
```

Default remains on. Not exposed in admin settings. Not group-based team-creation limits (#2698). Not per-team wizard skip (#2744).

## Checklist

- [x] Tests ([unit](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#unit-tests), [integration](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#integration-tests), api and/or acceptance) are included
  Admin toggle removed; no new UI surfaces
  Documented in PR description (`occ config:app:set`)
  no

## AI (if applicable)


lib/Service/TeamFolderPolicy.php
```diff
@@ -29,17 +31,35 @@
...
+	/**
+	 * Whether Circles may provision team folders (auto-create on team creation
+	 * and Circles UI/API upgrade). Controlled via config.php:
+	 *
+	 *   'circles.team_folder_auto_create' => false,
+	 *
+	 * Defaults to true when unset. OCC upgrade commands bypass this gate.
+	 */
+	public function isTeamFolderProvisioningEnabled(): bool {
+		$value = $this->config->getSystemValue(self::SYSTEM_TEAM_FOLDER_AUTO_CREATE, null);
```
artonge: Can't this be an app config instead?

artonge: I would try to avoid 'poluting' the system config for such specific configs

lennart441: It can be app config technically, but we put it in `config.php` on purpose.

This PR follows the direction from [@jospoortvliet on #2698](https://github.com/nextcloud/circles/pull/2698#issuecomment-5249432573): keep the strong default (teams + team spaces), and if a restriction is needed, only allow it in `config.php` so it is not the first thing admins reach for in the UI (and so we do not treat Nextcloud Teams like a feature to turn off). #2698 had this as an admin/app setting; we closed that PR in favor of this one for that reason.

App config would still be changeable via `occ config:app:set` and is easy to surface again in admin settings later. System config keeps this as an explicit ops-level opt-out, which matches that discussion.

Happy to revisit if you still prefer app config with the UI kept out.

artonge: Yes, let's keep it in the app config :)

artonge: To make sure that it is not miss-used.
```suggestion
	public function isTeamFolderAutoProvisioningEnabled(): bool {
		$value = $this->config->getSystemValue(self::SYSTEM_TEAM_FOLDER_AUTO_CREATE, null);
```

lib/Controller/TeamFolderController.php
```diff
@@ -52,8 +52,17 @@ public function getTeamFolder(string $circleId): DataResponse {
 
 	#[NoAdminRequired]
 	public function upgradeTeamFolder(string $circleId): DataResponse {
+		if (!$this->policy->isTeamFolderProvisioningEnabled()) {
```
artonge, line 55: The config is about blocking auto provisioning, but not blocking provisioning at all, no?

lib/Command/CirclesTeamFolderUpgrade.php
```diff
@@ -0,0 +1,141 @@
...
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Throwable;
+
+/**
+ * Admin escape hatch to provision team spaces for existing teams.
+ *
+ * This bypasses the config.php gate (`circles.team_folder_auto_create`) that
```
artonge: This is not a bypass if the config is only about blocking auto-provisioning.

artonge (commented):
This is still a system config and not an app config

lib/Service/TeamFolderPolicy.php
lennart441: Agreed the name is easy to misread, but `…AutoProvisioning…` would be the wrong contract.

This flag is meant to stop **user** provisioning (create + UI/API self-upgrade), not only auto-create, so quota stays under admin control. OCC is not gated.

Happy to rename to something like `isUserTeamFolderProvisioningEnabled` instead.

lib/Controller/TeamFolderController.php
lennart441: For this PR it is meant to block user provisioning, not only auto-create.

Leaving upgrade open would let anyone attach a space afterwards and consume quota. Admin provisioning stays on OCC. Optional skip per team is #2744.

lib/Command/CirclesTeamFolderUpgrade.php
lennart441: Fair on the wording. OCC is the admin path and must keep working when user provisioning is off. I will drop “bypass” in comments/docs and say: user paths honour the flag, OCC does not.

lennart441:
This PR is the original plan after #2698 / Jos: teams stay creatable, but user team-folder provisioning can be switched off so quota stays under admin control (OCC still works).

That is not really #2744 (optional skip per team in the wizard). Happy to open a dedicated issue and retarget if that helps.

lib/Service/TeamFolderPolicy.php
artonge: If this is not about auto-provisioning, let's rename the name of the config

lib/Controller/TeamFolderController.php
jospoortvliet: we discussed this: we don't care. We can block auto-provisioning, or provisioning at all - both are fine, so we are fine with @lennart441 do what is quickest for you.

lib/Command/CirclesTeamFolderUpgrade.php
jospoortvliet: see above - do what is quickest.

jospoortvliet:
> This PR is the original plan after #2698 / Jos: teams stay creatable, but user team-folder provisioning can be switched off so quota stays under admin control (OCC still works).
> 
> That is not really #2744 (optional skip per team in the wizard). Happy to open a dedicated issue and retarget if that helps.

yeah, two different things.  just discussed - so this doesn't have to be in config.php, just not visible in the admin UI - if it can be switched via the command line (occ or editing a config file), it's fine.  So to be configured with something like `occ app:config:system:set app_id --value "new_value"`.

But we agreed it is not a big deal either way, so @lennart441 do what is quickest for you:
* merge this as it is (so other PR's can continue and build on it) and do a separate PR to move this config to the app config
* change the PR to move this (back?) to app config and then merge.

So whatever works ;-)

lennart441:
Moved the flag from system/`config.php` to **app config** as discussed (@artonge / @jospoortvliet):

```bash
occ config:app:set circles team_folder_auto_create --value="false" --type=boolean
```

Still not in admin UI. User paths (auto-create + UI/API upgrade) honour it; `occ circles:team-folder:upgrade` does not.

lib/Command/CirclesTeamFolderUpgrade.php
artonge, line 1: I don't think we need that command.

lennart441: Just to clarify: do you mean only circles:team-folder:upgrade, or the whole PR (app-config kill switch for user provisioning)?

If it’s only the OCC command, I can drop it and keep the app-config flag. Note that with the flag off there would then be no admin path to provision spaces unless we leave UI/API upgrade open.

If you mean the whole approach isn’t needed anymore, I’ll close the PR — happy to confirm before I do that. (Context: this would be the second PR on this topic after #2698 that we implemented based on earlier feedback and then walked back.)

artonge: I only mean the command.

I don't see use cases to allow bypassing the app-config flag.

lennart441:
Dropped `circles:team-folder:upgrade` as requested — no bypass of the app-config flag.

lib/Command/CirclesTeamFolderUpgrade.php
lennart441: @artonge done

lib/Controller/PageController.php
artonge, line 1: Probably not needed anymore.

artonge (approved):
Good besides nitpick.

artonge:
And please cleanup your git history

lennart441:
> And please cleanup your git history

Squashed to a single commit on current master

src/teams/team-page/components/CircleDetails.vue
```diff
@@ -326,6 +326,8 @@ import { getTeamFolder, upgradeTeamFolder } from '../../api.ts'
 
 import 'cropperjs/dist/cropper.css'
 
+const teamFolderProvisioningEnabled = Boolean(loadState('circles', 'teamFolderProvisioningEnabled', true))
```
artonge, line 329: Not needed too.

lennart441: Thanks for the extra look.

The initial state / canCreateTeamFolder check is intentional, not leftover from the OCC command.

The goal is that admins can keep quota fully under their control. If team_folder_auto_create is off, users should not be able to consume team-space quota — whether the folder is created with the team or attached afterwards via the UI. Those two paths are the same from a quota perspective.

So we hide the upgrade action in the UI (and the API already returns 403) when the flag is off. Auto-create-only would still let someone create a team first and add the space later.

At least, that was my intention.

PR #2760, opened 2026-08-14
Add "New Deck board" shortcut

### LINKED ISSUE #2779: add "create Deck board" button to Teams page
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

## Feature request

We can (soon) create deck board owned by team. API/PR: https://github.com/nextcloud/deck/pull/8266

We should make sure the Teams UI supports it!

### LINKED ISSUE #8167: Allow a team (circle) to own a board
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

**Is your feature request related to a problem? Please describe.**
In a organization, people come and go. This can be a problem when they own resources that are actually used/relevant for a larger group of people. For Files, this was resolved via Teamfolders, and Collectives resolved it by always having a circle as owner. Deck should follow the same/a similar approach.

This is part of a wider effort to ensure team ownership is possible for all resources: https://github.com/nextcloud/circles/issues/2644

We have 3 types of roles in teams right now:
* owner (only person who can delete the team)
* moderators (can add/remove team members, rename team etc)
* normal members

We will add 'guests' (read-only members). After that, a rich role mechanism is on the agenda. This would including a custom role mechanism so complex permissions from a checklist ("can manage X") can be assigned for each role, then roles can be assigned to team members. Other settings we are working on is to dis-allow re-sharing on a team level, ideally Deck can respect that setting.

Much of that is future, but for Deck now the core issue team ownership should resolve is that no single person can block work or cause issues when unavailable or having left the company.

**Describe the solution you'd like**
It should be possible to designate a team as owner.

Based on the roles in teams that currently exist, a mapping like this would perhaps make sense:
* moderators can act as owner
* normal members can act as normal members
* guests have read-only access (future feature of Circles/Teams, so probably not for 35)

There are other questions and I thought of what I think are reasonable answers, but the designers might have different ideas of course:
* can all team members do everything? Are they all owners?
    * See mapping above. I think the team members should have rights that follow from their rights in the Team. That is, team managers should have the ability of an owner (able to change ownership, mainly), other members should have the access rights they have in the team. Currently, teams have no specific access rights, to that just means everything. In the future, teams will likely get at least an option to block re-sharing of resources from the team (which means the team this board is assigned to is the ONLY group of people who can have access - sharing should not be allowed). And we will probably introduce read-only team members.
        * this then adds the question: what happens to other current shares, esp if the new owning-team has a no-sharing policy. Remove or retain? I would suggest we warn the team managers about this in a notification, but retain the shares. But warning that sharing with the no-share team results in removing the other shares is also a legit approach... Maybe easier.
* What happens when a Team is removed?
    * just like with Deck boards of a user - delete. We will probably have to introduce an 'archive' function for teams, which should also archive their boards.
* What happens when Teams are not available?
    * you should not be able to transfer ownership to a team...
* What happens to you - can you still see the board, even if you're not in the team?
    * my suggestion is to add the user who does the ownership transfer to the board, unless they are in the team, and unless the team is has a no-sharing policy. Reason is that, in most cases, you were/are probably working on this task with others and now hand over ownership. If that is not the case, and should no longer have access, you can leave yourself, or an owner can remove you - this is not disruptive. But losing access right away until somebody re-adds you IS disruptive.

**Describe alternatives you've considered**
There could be a no-owner model, like Talk, where boards exist on their own... But I think this makes big orgs, who want tight ownership and clear responsibilities, very unhappy.

**Additional context**
I did an AI assisted PR earlier - but did not have the time to work on it more:
https://github.com/nextcloud/deck/pull/7899

samin-z:
* Resolves: [#8167](https://github.com/nextcloud/deck/issues/8167) from Deck app

## Summary
The API wiring of adding Decks to team.

## TODO

- [x] Call Deck api for creating new board.

## Checklist

- [ ] [Labels added](https://github.com/nextcloud/server/labels) where applicable (ex: bug/enhancement, `3. to review`, feature component)

## AI (if applicable)


artonge (commented):
It needs https://github.com/nextcloud/deck/pull/8266 right?

grnd-alt (approved):

grnd-alt (approved):

PR #2846, opened 2026-08-28
Design pass and team pages

marcoambrosini:
Re-organize the existing features, polishing and add a new team pages feature.

Screenshot 2026-08-28 at 17 23 16

Screenshot 2026-08-28 at 17 23 34

Screenshot 2026-08-28 at 17 23 45


lib/Controller/TeamTabsController.php
```diff
@@ -0,0 +1,131 @@
...
+class TeamTabsController extends OCSController {
+	private const SETTING_TAB_ORDER = 'teamsTabOrder';
+
+	/**
+	 * Structural bounds on the stored order. The ids are opaque to the
+	 * server; unknown ids are ignored by the client, so only their volume
+	 * needs limiting — the setting is loaded on every fetch of the circle
+	 * and replicated to federated instances.
+	 */
+	private const MAX_ID_LENGTH = 64;
```
artonge, line 51: Any reason for this limit?

LICENSES/LicenseRef-NextcloudTrademarks.txt
```diff
@@ -0,0 +1,9 @@
+The Nextcloud marks
+Nextcloud and the Nextcloud logo is a registered trademark of Nextcloud GmbH in Germany and/or other countries.
+These guidelines cover the following marks pertaining both to the product names and the logo: “Nextcloud”
+and the blue/white cloud logo with or without the word Nextcloud; the service “Nextcloud Enterprise”;
+and our products: “Nextcloud Files”; “Nextcloud Groupware” and “Nextcloud Talk”.
+This set of marks is collectively referred to as the “Nextcloud marks.”
+
+Use of Nextcloud logos and other marks is only permitted under the guidelines provided by the Nextcloud GmbH.
+A copy can be found at https://discord.com/branding
```
artonge: discord.com ?

I don't think we need this file.

src/teams/components/TeamAvatar.vue
```diff
@@ -3,12 +3,124 @@
...
+	}
+	// Concurrent consumers of the same circle share one request; a refresh
+	// always starts a fresh one to skip an in-flight stale response.
+	let request = refresh ? undefined : avatarRequests.get(circleId)
+	if (!request) {
+		request = fetchAvatarUrl(circleId, refresh).finally(() => avatarRequests.delete(circleId))
+		avatarRequests.set(circleId, request)
+	}
+	return request
+}
```
artonge: Let's drop caching avatar, this should be done by the browser.

vite.config.ts
artonge, line 1: Uneeded, let's revert.

REUSE.toml
```diff
@@ -76,3 +76,9 @@ path = ".l10nignore"
 precedence = "aggregate"
 SPDX-FileCopyrightText = "2026 Nextcloud GmbH and Nextcloud contributors"
 SPDX-License-Identifier = "AGPL-3.0-or-later"
+
+[[annotations]]
+path = ["img/dashboard/**.webp"]
+precedence = "aggregate"
+SPDX-FileCopyrightText = "Nextcloud GmbH <https://nextcloud.com/trademarks/>"
+SPDX-License-Identifier = "LicenseRef-NextcloudTrademarks"
```
artonge: Which licence are we usualy using for our assets?

src/teams/api.ts
```diff
@@ -293,6 +295,156 @@ export async function upgradeTeamFolder(teamId: string): Promise<TeamFolder> {
 	return data.ocs.data.folder
 }
 
+/**
+ * A team page: a markdown file stored in the root of the team folder,
+ * surfaced as a tab on the team (like the pinned tabs of a Microsoft
```
artonge: Maybe we want to drop the credit here ^^

src/teams/api.ts
```diff
@@ -293,6 +295,156 @@ export async function upgradeTeamFolder(teamId: string): Promise<TeamFolder> {
...
+/**
+ * Fetch the team-level tab order (readable by every member).
+ *
+ * @param teamId - The team single id
+ * @return Tab ids, first to last. Empty when no order has been saved.
+ */
+export async function fetchTabOrder(teamId: string): Promise<string[]> {
+	const { data } = await axios.get<OcsResponse<{ order: string[] }>>(
+		generateOcsUrl('apps/circles/teams/{circleId}/tab-order', { circleId: teamId }),
+		{ headers: HEADERS },
```
artonge: Headers are probably not needed here. They were dropped in a recently merged PR, a rebase should remove the variable anyway.

src/teams/api.ts
```diff
@@ -293,6 +295,156 @@ export async function upgradeTeamFolder(teamId: string): Promise<TeamFolder> {
...
+ *
+ * @param name - The collective name (the team's name)
+ */
+export async function createCollective(name: string): Promise<void> {
+	await axios.post(
+		generateOcsUrl('apps/collectives/api/v1.0/collectives'),
+		{ name },
+		{ headers: HEADERS },
+	)
+}
```
artonge, line 444: This should eventually not be hard coded in order to not reference the API of another app.

src/teams/views/PageView.vue
```diff
@@ -0,0 +1,171 @@
...
+	loading.value = true
+	error.value = null
+	destroyEditor()
+
+	const textApi = (window as unknown as { OCA?: { Text?: TextApi } }).OCA?.Text
+	if (!textApi?.createCollaborativeEditor) {
+		error.value = t('circles', 'The Text app is required to show team pages')
+		loading.value = false
+		return
+	}
```
artonge:
```suggestion
	if (window.OCA?.Text === undefined) {
		error.value = t('circles', 'The Text app is required to show team pages')
		loading.value = false
		return
	}
```

Then directly use `OCA.Text` instead of `textAPI`

artonge: Nitpick only to allow greping for OCA.Text.

src/teams/views/PageView.vue
```diff
@@ -0,0 +1,171 @@
...
+
+const resourcesStore = useTeamResourcesStore()
+
+const editorContainer = ref<HTMLElement | null>(null)
+const loading = ref(true)
+const error = ref<string | null>(null)
+
+let editor: TextEditorEmbed | null = null
+
+// The component instance is reused across page tabs, so loadPage runs can
```
artonge: Not sure that reusing the component is a good idea. It bloat the code, makes it more fragile, and I don't think that it brings that much performance improvments.

artonge (commented):
Good overall, still some comments.
I also would like to the opinion of a frontender before merging.

src/teams/views/PageView.vue
artonge: @pringelmann, this is a view injected by te router, is the reuse inevitable? Because it makes the code more complex than needed I think.

src/teams/resourcesStore.ts
```diff
@@ -0,0 +1,253 @@
...
+ * @param load - Performs the request and writes the slot
+ */
+function share(key: string, load: () => Promise<void>): Promise<void> {
+	let request = inFlight.get(key)
+	if (!request) {
+		request = load().finally(() => inFlight.delete(key))
+		inFlight.set(key, request)
+	}
+	return request
+}
```
artonge: Feels like premature optimisation, can the methods really be called multiple time in parallel?

lib/Controller/TeamTabsController.php
marcoambrosini: When a team admin drags the sidebar entries into a new order, the client sends the server a list of ids like ["team-folder", "page-42", "home"], and the server saves that list in the team's settings.
The problem the limits solve: the server just trusts whatever list it receives. Nothing in the browser UI stops someone from skipping the UI entirely and sending the request by hand, with a list of 10 million entries, or ids that are each a megabyte of garbage text. The server would happily save all of it.

LICENSES/LicenseRef-NextcloudTrademarks.txt
marcoambrosini: Oh this comes from the talk file!
https://github.com/nextcloud/spreed/blob/main/LICENSES/LicenseRef-NextcloudTrademarks.txt

src/teams/views/PageView.vue
marcoambrosini: A bit of background: we can have many text page tabs and users can quickly switch between them. So we need to make sure we keep track of whether editing sessions we create are valid. This is especially important when a user switches tab before an editing session is joined but when the promise for joining it is still hanging. 
Otherwise we could leak unused collaborative editing sessions for these files.

Example: 
1. User clicks page A → run # 1 starts (seq = 1, loadSeq = 1), starts awaiting editor creation.
2. User clicks page B before A finishes → watch fires, run # 2 starts (seq = 2, loadSeq = 2). Run # 1 is now stale but doesn't know it yet — it's asleep.
3. Run # 2 completes, mounts B's editor, sets loading = false. Screen shows page B. 
4. Run # 1's editor for page A finally resolves. Checkpoint: 1 !== 2 → it destroys the A-editor it just received and returns without touching a thing. Page B is undisturbed.

If you have a better idea here @pringelmann I can change this :)

pringelmann: Not inevitable. It's the router's default: `page/:fileId` is the same component at the same depth, so `page/42`  -> `page/43` swaps props and keeps the instance.

Keying the child router-view in `TeamPage.vue` fixes it:

```diff
-<router-view />
+<router-view :key="$route.path" />
```

Fresh instance per page, so the counter and all the `seq !== loadSeq` checks go, the watcher becomes `onMounted(loadPage)`, and the `destroyEditor()` at the top of `loadPage` is dead code. Sibling tabs are unaffected, they remount either way.

One guard still has to stay. Navigate away while `createCollaborativeEditor` is still running and `onBeforeUnmount` fires first, while `editor` is still null. The editor shows up afterwards and nothing destroys it.

One flag covers that: set `unmounted = true` in `onBeforeUnmount`, then `created.destroy()` instead of assigning it.

src/teams/resourcesStore.ts
pringelmann: They can, but only because the loading is in the wrong place. `TeamHomeView` and `TeamFolderView` each fetch the folder and resources themselves, on top of what the sidebar already loads for the team.

Better to fix that than to dedupe around it. Moving the load into `TeamPage`, the route that owns the team scope, and letting the sidebar and the views just read the store. Then this can go @marcoambrosini

src/teams/views/TeamHomeView.vue
```diff
@@ -0,0 +1,211 @@
...
+				</div>
+				<ul class="item-list">
+					<ListItem
+						v-for="resource in group.resources"
+						:key="resource.id"
+						:href="resource.url"
+						:name="resource.label">
+						<template #icon>
+							<!-- eslint-disable-next-line vue/no-v-html -->
+							<div v-if="resource.iconSvg" class="resource__icon" v-html="resource.iconSvg" />
```
pringelmann: Even if an exploit path is narrow we should try to avoid XSS vectors like this if the input cannot be fully trusted.

`NcIconSvgWrapper` is already imported here and sanitises properly:

	<NcIconSvgWrapper v-if="resource.iconSvg" :svg="resource.iconSvg" class="resource__icon" />

Kills the eslint-disable too

src/teams/components/TeamNavigation.vue
```diff
@@ -0,0 +1,571 @@
...
+					<NcIconSvgWrapper :path="entry.icon" :size="20" />
+				</template>
+
+				<!-- External entries (the collective) open outside the app -->
+				<template v-if="entry.href" #counter>
+					<NcIconSvgWrapper inline :path="mdiArrowTopRight" :size="16" />
+				</template>
+
+				<template v-if="entry.page && isTeamAdmin" #actions>
+					<NcActionButton closeAfterClick @click="onDeletePage(entry.page)">
```
pringelmann, line 561: Deletes on click, right under Rename in the same menu, on a doc the whole team shares. Trash makes it recoverable, but `useTeamActions` already does `showConfirmation` for Leave and Delete team. Maybe we do the same here?

src/teams/api.ts
```diff
@@ -293,6 +295,156 @@ export async function upgradeTeamFolder(teamId: string): Promise<TeamFolder> {
 	return data.ocs.data.folder
 }
 
+/**
+ * A team page: a markdown file stored in the root of the team folder,
+ * surfaced as a tab on the team (like the pinned tabs of a Microsoft
+ * Teams channel).
+ *
+ * TODO: store the pages in the team folder's .system folder instead of
```
pringelmann: Can we decide before this ships? Once people start making pages the files are in the folder root, and moving them to `.system` later becomes a migration.

pringelmann:
Really nice work, this is a big step up from what we had!

One process thing, not a blocker: this is quite a lot to review all at once. ~3.5k new lines over 40+ files with major changes covering four different code sites.

Not asking you to split it now, that rebase would be miserable. But maybe next time start with something more self-contained like the tab ordering (new controller, new route, one store field). This is easier for everyone as a standalone PR to review.

Github also just made it simpler to stack your PRs so you aren't blocked by reviews:
https://github.github.com/gh-stack/
https://docs.github.com/en/pull-requests/how-tos/stacked-pull-requests

src/teams/api.ts
pringelmann: Not a blocker :)

pringelmann (approved):
No blocking comments

lib/Controller/TeamTabsController.php
artonge: Then let's rephrase the comment to make it understandable by the lay person.

artonge (changes requested):

src/teams/api.ts
marcoambrosini: done

src/teams/router.ts
```diff
@@ -25,24 +28,60 @@ const routes: RouteRecordRaw[] = [
...
+				name: 'team-members',
+				path: 'members',
+				component: TeamMembersView,
+				props: true,
+			},
+			{
+				// The settings page shipped in earlier releases; its links
+				// land on the team, where settings is in the sidebar footer.
+				path: 'settings',
+				redirect: { name: 'team' },
```
artonge: We do not need to keeping backward compatibility of links, so you can drop that entry.

src/teams/api.ts
```diff
@@ -290,11 +277,241 @@ export async function upgradeTeamFolder(teamId: string): Promise<TeamFolder> {
...
 	)
 	return data.ocs.data.folder
 }
 
+/**
+ * Subfolder of the team folder holding the page files, nested inside the
+ * hidden `.system` folder so the pages neither clutter the files the team
+ * actually shares nor the `.system` root itself.
+ */
+const PAGES_FOLDER = '.system/team-pages'
```
artonge: I think we should adopt a scoped directory structure that uses the app name. Then the app owns whatever is within its directory.

```suggestion
const PAGES_FOLDER = '.system/teams/pages'
```

LICENSES/LicenseRef-NextcloudTrademarks.txt
artonge, line 1: Can you double-check that this is needed?

src/teams/api.ts
marcoambrosini: good point :)

LICENSES/LicenseRef-NextcloudTrademarks.txt
marcoambrosini: This is needed for the new teams logo in the home page, forgot to say earlier 
Screenshot 2026-09-02 at 12 35 39

artonge:
Psalm issue is not related, merging

PR #2936, opened 2026-09-22
fix(teams): hide circles the user is not a member of

marcoambrosini:
The `/circles` endpoint returns visible/public circles (with `initiator: null`) alongside the user's own. These were being shown on the Teams home page, causing 403 errors when the app tried to load details for them.

Filter circles with `initiator == null` out before mapping to `Team` objects.

> [!NOTE]
> AI-assisted: ClaudeCode:claude-sonnet-4-6


marcoambrosini:
@pringelmann codecov seems to be unrelated, or am I getting something wrong?

pringelmann:
> @pringelmann codecov seems to be unrelated, or am I getting something wrong?

It might be that the `?? []` branch is not tested. So adding a test case where `ocs.data ` is `unefined` will probably fix it. Or dropping the guard entirely if we know for a fact that it can never be undefined, although that's perhaps more risky.

pringelmann:
> > @pringelmann codecov seems to be unrelated, or am I getting something wrong?
> 
> It might be that the `?? []` branch is not tested. So adding a test case where `ocs.data ` is `unefined` will probably fix it. Or dropping the guard entirely if we know for a fact that it can never be undefined, although that's perhaps more risky.

Pushed an extra test that should fix it

pringelmann (approved):

marcoambrosini:
@pringelmann how do we backport this to the current version? I'm not familiar with how this works in circles

pringelmann:
@marcoambrosini using the same backport command as server. This one failed though, so I had to do it manually: https://github.com/nextcloud/circles/pull/2942

marcoambrosini:
Thanks @pringelmann!

